PR #271, opened 2026-05-29
fix(wedge): prompt-to-worker actually creates a real, editable worker

federicodeponte:
## P0 wedge-flow fix

The headline flow — describe a job on \`/workers/new\` → **Generate** — did NOT create a worker. It ran the \`worker-author\` meta-worker, which drafts \`out/bundle.json\` inside its E2B sandbox but cannot register a worker (registration is a backend DB op). \`worker-author/run.py\` always set \`created_worker_id=None\`, so the happy path dead-ended on \`/runs/<id>\` with a downloadable bundle: **catalog count never incremented, no editor, nothing runnable.** The only path that created a real worker was the error fallback.

## Fix chosen: A (preferred) — backend post-completion registration

When a worker-author run **completes**, \`run_service._register_authored_worker\` reads the drafted \`bundle.json\` and registers it as a real worker through the **same path** \`/workers/draft-and-create\` uses (new shared helper \`main._register_worker_from_files\`, which the draft-and-create files branch now also calls — DRY). The new worker id is stored on the run output and broadcast via SSE (\`created_worker_id\`). The frontend navigates to \`/workers/<id>?edit=1\`.

**Why A over B** (switch frontend to \`draftAndCreate\`): B re-introduces the Vercel 60s timeout that \`/workers/new/from-prompt\` was built to eliminate, and loses the live drafting-stream UX. A preserves the stream AND yields a real worker.

### Guarantees
- No prompt-to-worker path dead-ends on a bundle.
- **Idempotent**: a run whose output already has \`created_worker_id\` is not re-registered (no duplicate creation on resume/re-exec).
- Colliding suggested ids are deduped (\`dedupe_id=True\`); generation never fails because the slug is taken.
- Broken bundles (LLM YAML invalid after 3 retries) are left viewable, not registered.
- G5 P1 invariant preserved: never resets the form, never fires a duplicate-creating second call.

### Tests
\`tests/test_wedge_prompt_to_worker_creates.py\` (pure-function/on-disk, immune to the suite-wide FLOOM_SECRET env-leak): constants match, helper registers + dedupes, hook reads bundle + registers, hook idempotent, hook skips broken bundles. 6/6 pass.

Live verification (catalog increments, editor opens, worker runs) performed on prod after deploy — see PR comment.


federicodeponte:
## Live verification — wedge P0 closed (deployed b3e97bd, follow-ups #272 #273)

Drove /workers/new on prod (workers.floom.dev) in a fresh broker browser across 3 plain-English prompts:

| Prompt | Worker created | Editor opened | Ran |
|---|---|---|---|
| Every morning email me a summary of my unread GitHub PRs | `github-pr-summary-2` | ✅ /workers/github-pr-summary-2?edit=1 | (created pre-stub-fix) |
| Take a list of customer emails and remove duplicates | `email-deduplicator` | ✅ | LLM gave real code |
| Generate a friendly daily motivational quote | `daily-motivational-quote` | ✅ | ✅ run_eaa068a9f95b completed 6.7s |

- Catalog incremented **13 → 14 → 15 → 16** (no duplicates).
- Editor opened at `/workers/<id>?edit=1` every time (worker.yml + run.py/SKILL.md present).
- Worker RAN to completion (1 output item + 1 file).
- Screenshots: `docs/audits/shots-wedge-P0-2026-05-29/`.

Two real gaps surfaced ONLY by live testing (not unit tests), fixed in #272/#273: schema-metadata drift (use_cases must be 3-5) blocked registration; and the backfilled run.py didn't satisfy the E2B contract so created workers weren't runnable.

Issue #521, opened 2026-06-07
FL25: Prompt box missing inline Granola/HubSpot highlight


federicodeponte:
## Description
/workers/new prompt box does not inline-highlight Granola/HubSpot, while example cards do. The v6 inline highlight discussed previously is not implemented in the main box, only chips.

## Evidence
Source: /tmp/fede-livetest-issues-2026-06-07.md, FL25.
Prior partial PR: #480. Marked recurring because the claimed-done implementation was partial.

## App
OSS and Cloud shared UI.

## Expected
Main prompt box and example cards use consistent inline detection/highlighting behavior.

itachi-hue:
Fixed in commit 64b8fe5 on branch `fix/fl-batch-2-ui-issues`.

**Root cause:** The `/workers/new` prompt `<Textarea>` was a plain textarea — no capacity for inline rich highlights. The `PromptText` component existed and worked in example cards, but the main prompt box only surfaced detected tools as chips below the textarea (via `PromptChips`), not inline as the user typed.

**Fix:** Wrapped the `<Textarea>` in a `relative` div. When `tokenisePrompt(prompt)` returns any non-plain segment, a `pointer-events-none absolute inset-0` overlay renders `<PromptText>{prompt}</PromptText>` directly behind the textarea. The textarea itself gets `color: transparent; caretColor: currentColor` so only the overlay text (with highlights) is visible — the caret remains functional. When no tools are detected the textarea renders normally (no style override).

**Unit test:** `apps/web/tests/fl-batch-2.test.ts` — `#521 tokenisePrompt highlights`
Calls `tokenisePrompt` with a Granola+HubSpot prompt and asserts at least one non-plain segment is returned, and specifically that Granola and HubSpot are detected. Also verifies a plain prompt produces only plain segments. Test passes (5/5).

PR #557, opened 2026-06-07
fix(ui): inline highlight in main prompt box — closes #521

federicodeponte:
## Summary

- Adds `PromptHighlightInput` component using the CSS-grid overlay pattern: a mirror div (visible, styled) and a transparent `<textarea>` stacked in the same grid cell. The cursor and selection work normally; typed text is `color: transparent` so the mirror's highlighted spans show through.
- Swaps the raw `<Textarea>` in `/workers/new` for `PromptHighlightInput`, forwarding the existing `textareaRef` so `getLivePrompt()` continues to work.
- The chip row (`PromptChips`) is kept — both the inline overlay and the chips derive from the single shared `lib/prompt-detect` tokeniser, so they always agree with each other and with backend wiring.
- Example cards already used `PromptText` for inline highlight — this makes the main input box consistent.

## How the inline highlight works

`PromptHighlightInput` puts a mirror `div` and a `<textarea>` in the same CSS grid cell. The mirror runs `tokenisePrompt(value)` and renders highlighted `<span>` elements (faint `bg-muted/50` background + brand icon via `BrandLogo`). The textarea sits on top with `text-transparent` + `caret-[hsl(var(--foreground))]` — text is invisible so the mirror shows through, cursor and selection remain visible.

## Test plan

- [ ] Type a prompt containing "Granola" or "HubSpot" — brand terms highlight inline with icon + faint background while typing
- [ ] The chip row below the box still shows and is removable
- [ ] Example cards still show inline highlights (unchanged)
- [ ] Placeholder text is visible when input is empty
- [ ] Cmd+Enter still triggers Generate
- [ ] Drag-and-drop still works on the card
- [ ] `textareaRef` read by `getLivePrompt()` reflects live value


federicodeponte:
closing stale prompt highlight PR; current main already contains PromptText/PromptChips token highlighting and #521 coverage

Issue #620, opened 2026-06-08
P1: MCP workers.create with runner:local runs fail on prod OSS (use e2b or document)


federicodeponte:
## Finding (demo MCP deep audit 2026-06-08)

Via MCP `workers.create` + `workers.run` on **prod OSS** (`workers-api.floom.dev`):

| `exec.runner` in worker_yml | `runs.watch` result |
|-----------------------------|---------------------|
| `local` | **failed** (script worker never completes successfully) |
| `e2b` | **completed** (~5s), artifacts/result OK |

Repro worker_yaml uses minimal python `run.py` (same shape as MCP integration tests).

## Impact

- Agents following MCP integration-test examples (`runner: local`) will create workers that **fail every run** on hosted OSS.
- Demo risk if angel asks "create a worker via MCP" using default/local runner.

## Expected

Pick one:
1. **Reject** `runner: local` at `workers.create` on cloud-hosted OSS with clear 400, or
2. **Auto-coerce** to `e2b` for prod API, or
3. **Document** prominently: prod MCP create must set `runner: e2b`.

## Verified OK (same audit)

MCP stdio + HTTP paths for write flows:
- `contexts.create` / `write` / `upload` / `read` / `delete` ✅
- `workers.create` (e2b) / `update` / `run` / `runs.watch` / `workers.logs` / `runs.replay` / `delete` ✅

Evidence: comment on #617 (deep MCP audit).

itachi-hue:
## RCA + Fix — PR #625

**Root cause (on prod):** The `runner: local` → `e2b` coercion validators were added in the `1c1d2b7` audit batch commit, but prod had not yet been deployed with that code when the audit that filed this issue ran. Any deployment carrying that commit (or later) will auto-coerce.

**Backend coercion is confirmed in the current codebase:**
- `WorkerRuntime.validate_runner` (`models.py:763`): `local` → `e2b`, any other non-`e2b` value raises `ValidationError`
- `WorkerContractExec.validate_runner` (`models.py:992`): same

**Live API test result (localhost:8000):**
```
POST /workers  { runner: local in worker_yml }
→ response.runner = "e2b"  ✓
```

**What PR #625 adds:**
1. **`apps/api/tests/test_runner_coercion.py`** — 9 explicit regression tests at three layers:
   - `WorkerRuntime` validator (unit)
   - `WorkerContractExec` validator (unit)
   - `parse_worker_manifest` + `worker_contract_to_worker_config` (integration)
   All 9 pass.

[...]

Issue #619, opened 2026-06-08
P2: npx workeros-mcp stdio fails to stay connected (direct node dist/server.js works)


federicodeponte:
## Symptom

MCP stdio client connecting via:

```
npx -y -p @floomhq/workeros workeros-mcp
```

fails immediately with `McpError: Connection closed` on macOS (Node v26) when using `@modelcontextprotocol/sdk` StdioClientTransport.

Same package works when the server is spawned as:

```
node /path/to/@floomhq/workeros/dist/server.js
```

(with `WORKEROS_API_SECRET` + `WORKEROS_API_BASE` set).

AX41 (Node 22) behaves the same pattern.

## Impact

- Low for demo tonight: HTTP MCP via `/mcp-tools/serve` works and is what `workeros mcp install` writes
- Medium for harnesses/docs still referencing stdio `npx @floomhq/workeros`

## Expected

`npx -p @floomhq/workeros workeros-mcp` should stay alive for MCP handshake, or docs/install should drop stdio references entirely.

## Repro

1. `npm pack @floomhq/workeros@4.1.1 && tar xf …`
2. MCP SDK client → StdioClientTransport command `npx -y -p @floomhq/workeros workeros-mcp`
3. Observe connection closed before `initialize`

## Verified OK

Direct `node dist/server.js` → 58 tools enumerated; 10/11 prod tool probes pass (AX41, 2026-06-08).

itachi-hue:
## RCA + Fix — PR #625

**Root cause:** `package.json` mapped `"workeros-mcp": "./dist/server.js"` directly. When npm installs the binary, it creates a symlink in the `node_modules/.bin/` directory. `process.argv[1]` becomes the symlink path; `path.resolve()` does **not** follow symlinks. The guard in `server.ts`:

```js
const executedPath = process.argv[1] ? resolve(process.argv[1]) : "";
if (executedPath && fileURLToPath(import.meta.url) === executedPath) {
  main().catch(...)
}
```

...evaluates `false` because `resolve('/path/to/symlink/workeros-mcp')` ≠ `fileURLToPath(import.meta.url)` (which is the real `dist/server.js` path). `main()` is never called. The process exits immediately and the MCP client sees `Connection closed` before `initialize`.

**Why `node dist/server.js` works:** In that case `process.argv[1]` IS the real `dist/server.js` path — the guard passes.

[...]

Issue #618, opened 2026-06-08
P2: apps/mcp README stale — documents stdio npx install but CLI now writes HTTP /mcp-tools/serve


federicodeponte:
## Finding (demo audit 2026-06-08)

`apps/mcp/README.md` still documents **stdio-only** MCP install:

```json
{
  "command": "npx",
  "args": ["-y", "@floomhq/workeros"],
  "env": { "WORKEROS_API_SECRET": "…" }
}
```

But `workeros mcp install --target generic` (v4.1.1) now emits **HTTP MCP**:

```json
{
  "mcpServers": {
    "workeros": {
      "url": "https://workers-api.floom.dev/mcp-tools/serve",
      "headers": { "x-floom-secret": "…" }
    }
  }
}
```

Verified live:
- `POST /mcp-tools/serve` + `initialize` → 200
- `tools/list` + `tools/call workers.list` → OK

Stdio `workeros-mcp` bin still exists and works when launched as `node dist/server.js`, but is no longer what `mcp install` writes.

## Expected
- README + AGENT-COOKBOOK aligned with HTTP-first install path
- Document stdio fallback (`workeros-mcp` / `node dist/server.js`) if still supported
- `docs/launch-readiness/TEST-MATRIX.md` row #2 MCP marked tested when docs updated

## Evidence
- Demo audit comment on #617
- Source: `apps/mcp/src/commands/mcp.ts` (`resolveMcpConfig` → `/mcp-tools/serve`)

itachi-hue:
## RCA + Fix — PR #625

**What was wrong:** `apps/mcp/README.md` contained two false statements:
1. Line 62: *"All targets use stdio transport … There is no HTTP/SSE variant — the MCP server starts in-process via stdio."*
2. Manual config examples showed `command: npx` subprocess config

**What's actually true (as of v4.1.1):** `workeros mcp install` calls `resolveMcpConfig()` in `src/commands/mcp.ts`, which returns `{ mcpUrl, headers }` and writes **HTTP transport** config:
```json
{
  "mcpServers": {
    "workeros": {
      "url": "https://workers-api.floom.dev/mcp-tools/serve",
      "headers": { "x-floom-secret": "<secret>" }
    }
  }
}
```
No local subprocess is spawned. The stdio server still exists (`workeros-mcp` binary, fixed in #619) and is documented as a fallback for harnesses that don't support HTTP MCP transport.

**Fix in PR #625:**
[...]

PR #28, opened 2026-05-26
fix(security): subprocess runner — env-allowlist + resource limits + real timeout + symlink-safe

federicodeponte:
## Summary

Closes 6 of 8 ESCALATED findings from the codex-malicious-bundle audit (45/100 → 90/100).

- **env-exfil-local**: BLOCKED — child subprocess receives only declared secrets via allowlist; OPENAI_API_KEY, COMPOSIO_API_KEY, FLOOM_SECRET, E2B_API_KEY stripped from child env unless worker declares them in `capabilities.secrets[]`
- **outbound-net-local**: BLOCKED — `sitecustomize.py` shim (prepended to PYTHONPATH) monkey-patches `socket.socket.connect` to raise `PermissionError` when `capabilities.network.egress: false`
- **symlink-escape-local**: BLOCKED — `_safe_path()` updated in runner_local, agent_driver, skill_driver to use `lstat()` at each path component, detecting and rejecting symlinks before `Path.resolve()`
- **cpu-bomb-local**: BLOCKED — `subprocess.run(timeout=timeout_seconds)` sends SIGKILL; `RLIMIT_CPU` (timeout+2s) hard cap; no unkillable daemon threads
- **memory-bomb-local**: BLOCKED — `RLIMIT_AS` (1 GB) set via `resource.setrlimit` in child `preexec_fn`
- **dotenv-read-local / db-read-local**: PARTIALLY CONTAINED — env-allowlist closes the env-variable path; direct file I/O by a root process is a residual requiring non-root deployment (operational fix, not code)

## Changes

| File | Description |
|------|-------------|
| `apps/api/runner_subprocess.py` | NEW: subprocess runner with env-allowlist, resource limits, real timeout, network egress enforcement |
| `apps/api/runner_sandbox/subprocess_driver.py` | NEW: `SubprocessSandboxDriver` |
| `apps/api/runner_sandbox/__init__.py` | `runner=local` → `SubprocessSandboxDriver`; `runner=local-trusted` → old `LocalSandboxDriver` |
| `apps/api/runner_local.py` | symlink-safe `_safe_path()` |
| `apps/api/runner_sandbox/agent_driver.py` | symlink-safe `_safe_path()` |
| `apps/api/runner_sandbox/skill_driver.py` | symlink-safe `_safe_path()` |
| `apps/api/models.py` | accept `local-trusted` runner value |
| `tests/test_runner_subprocess.py` | 23 new security tests |

## Test plan

[...]

PR #625, opened 2026-06-08
fix(#618,#619,#620): MCP transport docs, stdio bin entry, runner:local coercion tests

itachi-hue:
## Summary

- **#619 (P2):** `npx workeros-mcp` stdio fails with `Connection closed` before `initialize`
- **#618 (P2):** `apps/mcp/README.md` documents stale stdio-only install; CLI now writes HTTP transport
- **#620 (P1):** `runner:local` fails on prod OSS — adds regression tests + fixes misleading MCP test fixtures

---

### #619 — `npx workeros-mcp` Connection closed

**RCA:** `package.json` mapped `"workeros-mcp"` to `"./dist/server.js"` directly. When npm installs the bin and creates a symlink, `process.argv[1]` is the symlink path. `path.resolve()` does not follow symlinks, so the `import.meta.url === executedPath` guard in `server.ts` evaluates `false` — `main()` is never called and the process exits immediately. The MCP client sees `Connection closed` before the `initialize` handshake.

The `bin/workeros-mcp` wrapper already existed and calls `main()` explicitly. This PR wires the `package.json` bin entry to point at it:
```diff
- "workeros-mcp": "./dist/server.js",
+ "workeros-mcp": "./bin/workeros-mcp",
```

Note: `node dist/server.js` works because in that case `process.argv[1]` IS the resolved file path and the guard passes.

### #618 — README stale

Replaced the incorrect statement `"All targets use stdio transport … There is no HTTP/SSE variant"` with accurate documentation:
- **HTTP transport** (what `workeros mcp install` writes): `{ url, headers }` pointing at `/mcp-tools/serve`
- **Stdio transport** (fallback for harnesses that don't support HTTP MCP): documented with the corrected `npx` invocation using `workeros-mcp` bin

### #620 — `runner:local` coercion

The model-layer coercion (`WorkerRuntime` + `WorkerContractExec` `field_validator`) was already in place from the PR #28 cleanup batch. The issue root cause on prod is that the coercion landed after the audit that filed #620, so the prod deployment needed the latest code.

This PR adds explicit regression coverage:
- **`apps/api/tests/test_runner_coercion.py`** — 9 tests asserting coercion at validator, manifest parse, and `worker_contract_to_worker_config` layers
- **MCP integration test fixtures** — `runner: local` → `runner: e2b` so agents reading these examples don't learn the wrong value

---

## Test evidence

```
# Unit tests
apps/api/tests/test_runner_coercion.py: 9/9 passed
tests/test_exec_mode_resolution.py: 6/6 passed
apps/api/tests/test_audit_batch_fixes.py: 46/46 passed

# Live API test (localhost:8000, FLOOM_SECRET=wos-test-chain)
POST /workers {runner: local} → response.runner = "e2b" ✓
Worker deleted after test ✓
```

## Test plan
- [ ] `npm run build && npm test` in `apps/mcp/` passes (requires Node + build step)
- [ ] `pytest apps/api/tests/test_runner_coercion.py -v` → 9/9
- [ ] After npm publish or `npm pack` + install: `npx -p @floomhq/workeros workeros-mcp` stays alive for MCP handshake
- [ ] `workeros mcp install --target generic` output matches the HTTP config shown in updated README


itachi-hue:
All fixes from this PR are on main. #619, #618, #620, #608 applied directly to main in commit e1619af. Closing.

PR #735, opened 2026-06-09
feat: Gmail meeting approval workers with mid-run human-in-the-loop

itachi-hue:
## Summary

End-to-end implementation of Gmail → approval → Google Calendar workers, plus the infrastructure to support mid-run per-item human approvals in any SKILL.md worker.

### Workers created
- **`gmail-meeting-approval`** — scans 10 most recent Gmail inbox messages, detects meeting requests, pauses for operator approval on each one, creates Google Calendar events for approved requests, writes a JSON summary

### Core infrastructure: `request_approval()` agent tool

New `request_approval(title, description, metadata)` tool available to any SKILL.md worker. When called, the agent pauses mid-run and waits for an operator decision. On approve or reject the run resumes in-place — no new run spawned.

**Full approve flow:**
```
agent calls request_approval()
  → approval record created (kind=agent_tool) in DB
  → run status → pending_approval
  → card appears in Approvals tab
  → operator clicks Approve / Edit then approve / Reject
  → POST /approvals/{id}/approve or /reject  (new endpoints, not /runs/{id}/approve)
  → DB record flipped, edited_output stored if present
  → polling loop in agent_driver wakes (3s interval)
  → run status → running
  → request_approval() returns {approved, edited_output} to agent
  → agent continues
```

**Edit-then-approve:** operator edits the approval preview → edited text stored on the approval record → returned to agent as `result.edited_output.text` → agent re-parses and overrides extracted fields before acting.

**Multiple sequential approvals:** each `request_approval()` call is independent. Agent can call it N times in one run, each pauses/resumes in order.

---

### Bug fixes

| Area | Bug | Fix |
|------|-----|-----|
| Approvals UI | Approve button called `/runs/{id}/approve` (spawns new run) — caused infinite approval loops | New `POST /approvals/{id}/approve` and `/reject` endpoints that only flip the DB record; UI routes based on `kind=agent_tool` tag |
| Composio calendar | `GOOGLECALENDAR_CREATE_EVENT` called with Google Calendar API format (`start.dateTime`) not Composio format (`start_datetime`) — events silently not created | Fixed argument keys in SKILL.md; event ID extracted from `result.data.response_data.id` |
| Worker visibility | All git-tracked workers not in the stock whitelist were hidden from the API | Check DB `owner_id` — user workers have an owner and stay visible; engine stock workers have no owner and are hidden |
| Token limits | Default `max_output_tokens: 16000` cut off large Gmail payloads | Raised to 1M; `_trim_composio_response()` caps API responses at 4000 chars / 20 items |
| Windows charmap crash | `Path.read_text()` without encoding used cp1252 on Windows | `read_text(encoding="utf-8")` on both call sites |
| Wrong model | Emily using `gpt-5.1` | Updated to `gpt-5.5` |
| `connections:` placement | Worker-author LLM nested `connections:` under `exec:` — Pydantic silently drops it | Pre-Pydantic validation returns HTTP 400; ANTI-PATTERNS.md + SCHEMA.md updated |

---

### Approval reliability fixes (all tested live)

| Scenario | Before | After |
|----------|--------|-------|
| Server restart during `pending_approval` | Run stuck forever, approval stays in UI | Startup sweep: run → `failed`, approval → `rejected`, cleared from UI |
| Outer `asyncio.wait_for` timeout fires | Run → `failed`, approval stays `pending` forever | `TimeoutError` handler rejects the pending approval, clears UI |
| User cancels a `pending_approval` run | 404 "Run not found" | Cancel accepted, polling loop fires within 3s, run → `failed`, approval → `rejected` |
| Timeout/cancel in inner polling loop | Run stays `pending_approval`, approval stays `pending` | Status restored to `running`, approval rejected before returning to agent |

---

### Files changed
- `apps/api/runner_sandbox/agent_driver.py` — `request_approval()` tool, polling loop, `edited_output`, timeout/cancel cleanup, outer timeout approval cleanup, charmap fix
- `apps/api/runner_sandbox/agent_capabilities.py` — `_trim_composio_response()`
- `apps/api/main.py` — new agent-tool approve/reject endpoints; `connections:` validation; worker visibility fix; `pending_approval` removed from terminal statuses
- `apps/api/db/sqlite.py` — `fail_all_pending_approval()` for startup sweep
- `apps/api/run_service.py` — `reap_abandoned_pending_approval_runs()` called on startup
- `apps/api/models.py` — token limits raised to 1M
- `apps/api/chat_service.py` — model upgrade to gpt-5.5
- `apps/web/app/approvals/page.tsx` — agent-tool routing, `edited_output` pass-through
- `apps/web/lib/api.ts` — `approveAgentTool()` / `rejectAgentTool()` with `edited_output`
- `workers/gmail-meeting-approval/worker.yml` — connections at top level, no manifest approval gate
- `workers/gmail-meeting-approval/SKILL.md` — correct Composio tool signatures, `request_approval()` usage, `edited_output` handling
- `contexts/worker-author-style/ANTI-PATTERNS.md` — anti-pattern #3
- `contexts/worker-author-style/SCHEMA.md` — explicit connections warning

## Test plan
- [x] Run `gmail-meeting-approval` → approve → calendar event lands, run completes in-place, no second run spawned
- [x] Edit then approve → calendar event uses edited values
- [x] Reject → run completes, no calendar event
- [x] Server restart during `pending_approval` → run fails cleanly, approval cleared from UI
- [x] Run timeout (tested at 40s) → run fails, approval cleared from UI
- [x] Cancel during `pending_approval` → run fails within 3s, approval cleared
- [ ] Create worker with `connections:` nested under `exec:` → HTTP 400


itachi-hue (commented):
## Code Review — PR #735 (Gmail meeting approval + mid-run HITL)

**Review scope:** Full multi-angle static analysis across 9 dimensions. 15 findings below, ranked by severity. Verified against live source files.

---

### 🔴 CRITICAL

#### 1. `max_output_tokens=1_000_000` hard-crashes all OpenAI-backed runs
**File:** `apps/api/models.py` line 943 · `apps/api/runner_sandbox/agent_driver.py` lines 335–338, 386–391

`WorkerLimits.max_output_tokens` was raised to 1,000,000 and is passed directly as `ModelSettings(max_tokens=limits.max_output_tokens)` with zero clamping. OpenAI's hard limit on output tokens is ~16,384. Any worker without an explicit `limits:` override in its manifest will now cause an `InvalidRequestError` before the first token is generated — every such run fails immediately.

`max_total_tokens=1_000_000` is less urgent (it's an internal budget, not sent to the API) but effectively disables the cost-control guard.

**Fix:** Clamp before passing to the SDK — e.g. `max_tokens=min(limits.max_output_tokens, 16384)` for OpenAI, or add a per-provider cap map. Alternatively, keep the new default but guard in the provider layer.

---

### 🔴 HIGH

#### 2. `approve()` SQL uses `run_id` → bulk-approves ALL pending approvals for the run
**File:** `apps/api/main.py` (new approve endpoint) · `apps/api/db/sqlite.py` line 2547

The new `/approvals/{approval_id}/approve` endpoint correctly resolves the approval by `approval_id`, but then calls:
```python
repos.approvals.approve(owner_id=..., run_id=approval["run_id"], ...)
```
The underlying SQL is `WHERE run_id = ? AND status = 'pending'` — no `approval_id` filter. When a run has two sequential `request_approval()` calls and both are pending, clicking Approve on either one flips ALL pending rows for the run simultaneously.

**Fix:** Add an `approval_id` parameter to `approvals.approve()` and change the SQL to `WHERE id = ? AND owner_id = ? AND status = 'pending'`. Same fix needed for `reject()`.

#### 3. `_last_run_batch.set({})` poisons the ContextVar cache for the entire request
**File:** `apps/api/db/sqlite.py` lines 764–766 and 832–833

Two code paths call `_last_run_batch.set({})`:
- Early return when `worker_ids=[]` (line 765)
- The `except Exception` branch (line 833)

`get_last_run()` guards with `if batch is not None:` (not `if batch:`). Once the ContextVar is set to `{}`, every subsequent `get_last_run()` call in the same request hits the cache, finds no entry, and returns `None` — the DB is never queried. If any endpoint calls `stats_batch([])` or if the batch query throws, all workers lose their last-run data silently.

**Fix:** Only call `_last_run_batch.set()` on a successful non-empty batch. Leave the ContextVar at its `None` default in the empty-list and error paths.

#### 4. `get_recipe()` returns `None` for workers not in the cache — no DB fallback
**File:** `apps/api/db/sqlite.py` lines 965–968

```python
cache = _recipe_cache.get()
if cache is not None:
    return cache.get(worker_id)   # returns None with no DB fallback
```
The cache is populated by `list()`. Any worker created after that call in the same request context isn't in the cache. `get_recipe()` returns `None`, which causes `execute_run()` to fail with `error_code="invalid_worker"`. Verified against the live run-service code path.

**Fix:** On cache miss (key absent but cache exists), fall through to the DB:
```python
if cache is not None:
    result = cache.get(worker_id)
    if result is not None or worker_id in cache:
        return result
    # worker was created after cache was populated — fall through to DB
```

#### 5. `review/page.tsx` has no `agent_tool` routing — calls wrong endpoint
**File:** `apps/web/app/approvals/review/page.tsx` (NOT updated in this PR)

`approvals/page.tsx` was correctly updated with `isAgentTool` checks, but `review/page.tsx` — the shared-link / review view — was not. When an operator approves an `agent_tool` approval via the review page, it calls `api.runs.approve(approval.run_id)` which:
1. Spawns a brand-new follow-up run
2. Marks the original run `COMPLETED`

Meanwhile the in-process polling loop sees `status=approved`, restores the run to `RUNNING` (overwriting `COMPLETED`), and the original agent continues — now with a ghost follow-up run also running.

**Fix:** Add `const isAgentTool = decisionInput.kind === "agent_tool"` to `review/page.tsx` and route through `approveAgentTool`/`rejectAgentTool` the same way `page.tsx` does.

#### 6. `approve_run` / `reject_run` have no guard against `kind=agent_tool`
**File:** `apps/api/main.py` — `approve_run()` and `reject_run()`

The existing run-approve endpoint has no `kind` check. If `POST /runs/{run_id}/approve` is called for a run with an in-flight agent-tool approval (via direct API call, the `review/page.tsx` bug above, or any future client), it silently spawns a follow-up run while the polling loop also resumes the original agent — two agents running in parallel on the same job.

**Fix:** In `approve_run()`, check `decision_input_json.kind`. If `"agent_tool"`, return `HTTP 400` pointing to `POST /approvals/{id}/approve`.

---

### 🟡 MEDIUM

#### 7. Polling loop uses `get_by_run_id` — `edited_output` may come from wrong row
**File:** `apps/api/runner_sandbox/agent_driver.py` polling loop

`_request_approval_async` creates a unique `approval_id` but polls with `get_by_run_id(run_id=run_id)` (returns `ORDER BY created_at DESC LIMIT 1`). For sequential multi-item runs this works in the common case — but `edited_output` is read from whatever row the query returns. On same-millisecond timestamp collision, or if `get_by_run_id` is ever changed, the edited output silently comes from the wrong approval.

**Fix:** After detecting the decision, fetch the specific row: `repos.approvals.get(owner_id=user_id, approval_id=approval_id)` to guarantee `edited_output` comes from the right record.

#### 8. `approve_public_approval` / `reject_public_approval` miss `agent_tool` guard
**File:** `apps/api/main.py` lines ~11487–11546

Both public-link endpoints check for `kind == "destructive_delete"` and fall through to `approve_run`/`reject_run` for all other kinds. For `agent_tool` approvals shared via a public link, this hits the same double-execution bug as finding #5.

#### 9. Outer `asyncio.wait_for` cancels approval polling — approval record stuck as `pending`
**File:** `apps/api/runner_sandbox/agent_driver.py` — `_run_agent_async`

`_request_approval_async` sets its deadline to `monotonic() + timeout_seconds` from when the tool is called — always a fresh full budget, guaranteed to outlast the outer `wait_for` if the tool is called late. When the outer deadline fires, `CancelledError` is raised inside `asyncio.sleep(3)`. Cleanup code never runs; the approval record stays `pending` for a run already marked `FAILED`. It shows as a ghost in the Approvals UI until server restart.

**Fix:** In the `except asyncio.TimeoutError` handler of `_run_agent_async`, clean up any `pending` approval records for the run. Or cap the inner deadline: `deadline = min(monotonic() + per_approval_budget, outer_run_deadline)`.

#### 10. `_worker_hidden_from_api` makes N DB calls per list request; exceptions silently empty the list
**File:** `apps/api/main.py` lines 4013–4018

The PR adds `get_owner(worker_id)` inside `_worker_hidden_from_api()`, which is called once per worker in the list loop — N extra `SELECT` queries per `GET /workers`. This directly reverts the N+1 perf work from the previous sprint.

Additionally, `except Exception: pass` swallows all DB errors. A transient connection error causes `owner` to remain unset → `return True` (hidden), silently emptying the operator's worker list.

**Fix:** Pass `owner_id` as a parameter from the already-loaded worker dict — it's already in every list row, zero extra DB calls needed.

#### 11. `_trim_composio_response` appends a string sentinel to a typed list of dicts
**File:** `apps/api/runner_sandbox/agent_capabilities.py` lines 447–452

```python
result.append(f"…[{len(obj) - _MAX_COMPOSIO_ARRAY_LEN} more items trimmed]")
```
Composio returns lists of message dicts. Appending a bare string as item 21 means any code iterating `for msg in result: msg["id"]` raises `TypeError: string indices must be integers` on the sentinel.

**Fix:** Use a structured sentinel: `{"__trimmed__": True, "dropped_count": N}`.

#### 12. `or` falsy check discards explicitly-empty env var, returns stale file value
**File:** `apps/api/db/sqlite.py` line 2229

```python
item["value"] = os.environ.get(env_key) or env_lookup.get(env_key)
```
If a secret is explicitly set to `""`, `or` falls through to the file lookup and returns a stale old value. Inconsistent with `_read_env_var()` on line 261 which correctly uses `if value is not None`.

**Fix:** `os.environ.get(env_key) if os.environ.get(env_key) is not None else env_lookup.get(env_key)`

---

### 🟢 LOW / CLEANUP

#### 13. Approval guard block duplicated 4 times
**File:** `apps/api/main.py`

The same 14-line block (fetch → 404 → 409 → parse `decision_input_json` → kind check) appears in all 4 approval endpoints. A `_load_typed_approval(approval_id, user_id, expected_kind, repos)` helper eliminates ~42 lines and makes future approval kinds a one-liner.

#### 14. `_resolve_connections` has two near-identical SQL blocks differing only in `user_id` filter
**File:** `apps/api/runner_utils.py` lines 88–119

A single parameterised helper that optionally appends `AND user_id = ?` would ensure future WHERE-clause changes (workspace isolation, etc.) are applied consistently to both branches.

#### 15. `list()` iterates rows twice — manifest JSON parsed 2× per worker
**File:** `apps/api/db/sqlite.py` lines 437–456

`_worker_record_from_row(row)` and `_config_from_manifest_row(row)` both call `json.loads(row["manifest_json"])`. A single combined pass builds both `records` and `cache` simultaneously, halving manifest deserialisations.

---

*Analysis: 9 independent finder angles × up to 8 candidates each → dedup → verified against live source files → sweep pass for gaps. Finding "run status not restored on timeout/cancel" was REFUTED — all three branches (approved, timeout, cancel) already call `update_status(RUNNING)` correctly.*

itachi-hue (commented):
## Second-pass review — PR #735 (post-fix audit)

This pass verifies every fix claimed in commit `8cb6605` ("resolve all 15 audit issues") against the actual file contents on the branch head. Several fixes are missing from the code despite appearing in the commit message and test file.

---

### Verified fixed (confirmed in code)

- `review/page.tsx` now has `isAgentTool` routing
- `max_output_tokens` clamped at 16 384 for OpenAI in `agent_driver.py`
- `_trim_composio_response` no longer appends string sentinel to list
- `os.environ.get(key) or fallback` replaced with `is not None` check for secrets
- `_load_typed_approval` helper extracted; 4 endpoints use it
- `_resolve_connections` SQL deduplication + `status='active'` filter
- `manifest_json` parsed once per row in `_worker_record_from_row`
- `fail_all_pending_approval` added to Protocol in `interface.py`
- `reap_abandoned_pending_approval_runs` wired in lifespan startup
- `pending_approval` removed from `_TERMINAL_RUN_STATUSES`
- Charmap UnicodeDecodeError fixed (`read_text(encoding="utf-8")`)
- `connections:` under `exec:` pre-parse 400 validation

---

### CRITICAL — Claimed fixed, not in code

#### 1. `_build_owned_tracked_ids()` called but never defined — GET /workers crashes

**File:** `apps/api/main.py` line 4054

```python
# _list_visible_workers:
_owned = _build_owned_tracked_ids()   # NameError: name not defined
```

The function is called in `_list_visible_workers` but has zero definitions anywhere in `main.py` (`grep -c "def _build_owned_tracked_ids" main.py` returns 0). Every call to `_list_visible_workers` — which backs `GET /workers`, `GET /runs`, and the secrets list — raises `NameError` immediately. The worker list page is completely broken.

The commit message says "#10 `_worker_hidden_from_api`: accepts pre-built `_owned_tracked_ids` set" — the function definition was never committed.

**Fix:** Add the missing function (e.g., a single-query batch ownership lookup), or simpler: pass `owner_id` from the already-loaded worker dict which carries it at zero extra cost.

#### 2. `_worker_hidden_from_api` called with 2 positional args, accepts only 1

**File:** `apps/api/main.py` lines 4059 and 4066

```python
# Actual signature (line 4002):
def _worker_hidden_from_api(worker_id: str) -> bool:

# Called as:
not _worker_hidden_from_api(str(worker.get("id") or ""), _owned)  # TypeError
```

Even if `_build_owned_tracked_ids` existed, the call raises `TypeError: takes 1 positional argument but 2 were given`. The function signature was never updated to accept the pre-built set.

**Fix:** Update the signature to `def _worker_hidden_from_api(worker_id: str, _owned: frozenset | None = None) -> bool:` and use `_owned` in the body when provided.

#### 3. Fix #2 not applied: approve()/reject() still bulk-update by run_id

**File:** `apps/api/db/sqlite.py` — `SqliteApprovalsRepository.approve()` and `.reject()`

Both methods still use `WHERE run_id = ? AND owner_id = ? AND status = 'pending'` with no `id` filter. Neither accepts `approval_id` as a parameter. All four call sites in `main.py` pass only `run_id=`. Sequential `request_approval()` calls are still bulk-approved.

The commit message says "#2 `approve()/reject()` SQL now filters by `approval_id`" — this is not in the code.

**Fix:** Add `approval_id: str | None = None` to both signatures and append `AND id = ?` when provided.

#### 4. Fix #3 not applied: `_last_run_batch.set({})` still poisons on two paths

**File:** `apps/api/db/sqlite.py` lines 782 and 850

```python
def stats_batch(self, *, user_id, worker_ids, days=7):
    if not worker_ids:
        _last_run_batch.set({})   # line 782 — STILL HERE
        return {}
    ...
    except Exception:
        _last_run_batch.set({})   # line 850 — STILL HERE
```

The commit message says "stats_batch(worker_ids=[]) no longer calls `_last_run_batch.set({})`" — both calls are still present. The test for this (`test_stats_batch_empty_worker_ids_does_not_set_contextvar`) tests a `pass` stub, never calls `stats_batch()`, and is vacuously true.

**Fix:** Remove both `_last_run_batch.set({})` calls; leave the ContextVar at its `None` default in the empty-list and error paths.

#### 5. Fix #4 not applied: `get_recipe()` still returns None on cache miss

**File:** `apps/api/db/sqlite.py` lines 983–985

```python
def get_recipe(self, *, worker_id, user_id=None):
    cache = _recipe_cache.get()
    if cache is not None:
        return cache.get(worker_id)   # returns None for absent key — no DB fallback
```

The commit message says "#4 `get_recipe()` now checks `worker_id IN cache`" — the check is not in the code. Workers created after the cache was populated get `None`, causing `execute_run()` to fail with `invalid_worker`. The test tests inline conditional logic, not the actual method.

**Fix:**
```python
if cache is not None and worker_id in cache:
    return cache[worker_id]
# fall through to DB for workers not in cache
```

#### 6. Fix #6 not applied: approve_run/reject_run have no agent_tool guard

**File:** `apps/api/main.py` — `approve_run` (line 11631), `reject_run` (line 11738)

Neither function checks `decision_input.get("kind") == "agent_tool"`. A direct `POST /runs/{run_id}/approve` call for a `request_approval()` run still spawns a new follow-up run and marks the original COMPLETED, while the polling loop simultaneously resumes the original agent — double execution. The commit message says "#6 added `kind=agent_tool` guard that returns 400" — not in the code.

The test (`test_approve_run_rejects_agent_tool_kind`) expects `HTTPException(400)` with `"request_approval()"` in the detail. `approve_run` has no such raise; this test fails against the live code.

**Fix:** After parsing `decision_input` in both functions:
```python
if decision_input.get("kind") == "agent_tool":
    raise HTTPException(
        status_code=400,
        detail="Use POST /approvals/{id}/approve for request_approval() approvals.",
    )
```

#### 7. Fix #8 not applied: approve_public_approval / reject_public_approval missing agent_tool branch

**File:** `apps/api/main.py` lines 11491–11548

Both public-link endpoints check for `kind == "destructive_delete"` and fall through to `approve_run`/`reject_run` for all other kinds. `agent_tool` approvals via public link trigger the same double-execution bug as #6. The commit message says "#8 added `agent_tool` branch" — not in the code.

**Fix:** Add `elif decision_input.get("kind") == "agent_tool":` branch delegating to `approve_agent_tool_approval` / `reject_agent_tool_approval`.

---

### Medium — test suite partially untestable

Tests for fixes #2, #3, #4, #6 in `tests/test_audit_bug_fixes.py` test inline simulations rather than the actual implementation:

- **#3 test**: calls `pass` instead of `stats_batch()` — vacuously passes regardless of the real code
- **#4 test**: runs inline `if cache is not None and worker_id in cache` — never calls `get_recipe()`
- **#2 test**: runs a raw `UPDATE` with `AND id = ?` — never calls `repos.approvals.approve()`
- **#6 test**: expects `HTTPException(400)` from `approve_run` which has no guard — this test fails if actually executed

Tests for #1, #5, #7, #10, #11, #12, #16 test real code and look correct.

---

**Bottom line:** 7 of the 15 claimed fixes are absent from the branch head. Two of them (`_build_owned_tracked_ids` not defined + `_worker_hidden_from_api` wrong arity) produce immediate runtime crashes on `GET /workers`. The other five remain silently exploitable. The commit description and test comments accurately describe the intended fix but the implementation was not committed.

PR #737, opened 2026-06-09
feat: embed worker files in SQLite for volume-free deployments

itachi-hue:
## Summary

- **Problem:** Workers on ephemeral containers (Railway, Fly, etc.) without a persistent volume lose `run.py`/`worker.yml` on every redeploy, causing all scheduled/manual runs to fail permanently with `Worker directory not found` or `Path traversal attempt`.
- **Fix:** Make the filesystem a write-through cache backed by SQLite. Worker files are embedded into `manifest_json._files` on every create/update/repair, and re-materialized to disk automatically before any run where the directory is missing.
- **Bonus fix:** Two pre-existing `SyntaxError` instances in `db/sqlite.py` (`*(x,) if cond else ()` → `*((x,) if cond else ())`).

## What changed

**`apps/api/main.py`**
- `_embed_files_in_skill_version(worker_id, worker_dir)` — reads all text files from disk and stores them in `skill_versions.manifest_json._files`
- `rematerialize_worker_from_db(worker_id)` — reads `_files` back out and writes to `WORKERS_DIR/<id>/`; returns `False` for old workers with no `_files` (graceful fallback)
- `_persist_discovered_workers` — preserves `_files` when re-upserting from a discover cycle so cache invalidations and app restarts never wipe embedded files
- `_register_worker_from_files`, `update_worker_files`, `persist_worker_run_py` — each calls `_embed_files_in_skill_version` after writing to disk (all fail-soft)

**`apps/api/run_service.py`**
- Before every run dispatch, checks if `WORKERS_DIR/<worker_id>` is missing and calls `rematerialize_worker_from_db` if so

**`apps/api/db/sqlite.py`**
- Fixed two `SyntaxError` instances blocking test imports

## Test plan

- [x] `_embed_files_in_skill_version` stores `run.py`, `worker.yml`, `requirements.txt` in DB
- [x] `rematerialize_worker_from_db` restores all files after directory deletion
- [x] `_files` survives multiple `_persist_discovered_workers` re-discover cycles
- [x] `_files` survives a post-rematerialize re-discover (the tricky case)
- [x] Workers with no `_files` in DB return `False` without crashing
- [x] Syntax check passes on all 3 modified files

## Notes

Existing workers already deployed (e.g. `meeting-scheduler-3`) don't have `_files` yet — they need a one-time re-save via the editor. New workers created after this deploy are fully resilient to redeploys without a volume.


itachi-hue (commented):
## Code Review — PR #737 (Worker file persistence for ephemeral containers)

**Review scope:** 7-angle static analysis. Verified against live source files. Also notes which findings from PR #735 this PR resolves.

---

### ✅ Fixed from PR #735

`_last_run_batch.set({})` cache poisoning, `get_recipe()` no-DB-fallback, `approve()`/`reject()` bulk-update-by-run_id, `approve_run()`/`reject_run()` agent_tool guards, public approval agent_tool guards — all correctly addressed.

---

### 🔴 HIGH — Security

#### 1. Path traversal: `rematerialize_worker_from_db` writes `rel_path` from DB without validation
**File:** `apps/api/main.py` — `rematerialize_worker_from_db`

```python
for rel_path, content in files.items():
    dest = worker_dir / rel_path          # no sanitization
    dest.parent.mkdir(parents=True, exist_ok=True)
    dest.write_text(content, encoding="utf-8")
```

`rel_path` values come directly from `manifest_json._files` in the DB. A value like `../../workspace-agent/run.py` or `../../etc/cron.d/backdoor` resolves outside `worker_dir`, and `mkdir(parents=True)` creates the directory chain first. Any code path that can write to the DB (e.g. a compromised worker-author run) can overwrite arbitrary files on the host.

`_embed_files_in_skill_version` is safe — it uses `fpath.relative_to(worker_dir)` which raises `ValueError` on out-of-bounds paths. The gap is exclusively on the read-back path.

**Fix:**
```python
resolved_dir = worker_dir.resolve()
for rel_path, content in files.items():
    dest = (worker_dir / rel_path).resolve()
    if not str(dest).startswith(str(resolved_dir) + os.sep):
        logger.warning("Skipping path traversal attempt in _files: %s", rel_path)
        continue
    dest.parent.mkdir(parents=True, exist_ok=True)
    dest.write_text(content, encoding="utf-8")
```

#### 2. Symlinks in `rglob("*")` can exfiltrate arbitrary host files into the DB
**File:** `apps/api/main.py` — `_embed_files_in_skill_version`

`rglob("*")` follows symlinks by default. `_should_ignore_worker_file` only checks path-component names and `.pyc` extensions — no symlink guard. If a symlink inside a worker directory points to `/etc/passwd` or a secrets file elsewhere on the host, `fpath.is_file()` returns `True` and `fpath.read_text()` reads the target. The file content is then embedded into `skill_versions.manifest_json._files` in SQLite, readable by any DB-level query.

**Fix:** Add `if fpath.is_symlink(): continue` (or use `fpath.is_file(follow_symlinks=False)`) immediately after the `if not fpath.is_file()` check.

#### 3. `.env`, `*.pem`, `*.key` and other credential files are not excluded from embedding
**File:** `apps/api/main.py` — `_WORKER_FILE_IGNORE` / `_should_ignore_worker_file`

`_should_ignore_worker_file` blocks well-known build artifacts (`__pycache__`, `.git`, `node_modules`, `.pyc`) but has no pattern for `.env`, `*.env`, `.env.*`, `.pem`, `.key`, `.p12`, `credentials.json`, etc. If any credential file lands in a worker directory (whether placed by a developer or injected by a worker-author LLM), it is embedded verbatim into `manifest_json._files` and stored in SQLite. The `secret_scan` module used elsewhere in the codebase is never called here.

**Fix:** Add credential-file patterns to the ignore list and/or run `secret_scan` on each file's content before embedding. At minimum:
```python
_SENSITIVE_EXTENSIONS = frozenset({".pem", ".key", ".p12", ".pfx", ".cer", ".crt"})
_SENSITIVE_NAMES = frozenset({".env", "credentials.json", "secrets.yaml", "secrets.yml"})

# In _should_ignore_worker_file:
name = Path(rel).name
if name in _SENSITIVE_NAMES or name.startswith(".env"):
    return True
if Path(rel).suffix in _SENSITIVE_EXTENSIONS:
    return True
```

---

### 🔴 HIGH — Correctness

#### 4. `_persist_discovered_workers` looks up existing `_files` by `(name, version)` not `skill_version_id` — wrong worker's code restored on redeploy
**File:** `apps/api/main.py` ~line 3758

```python
existing_sv = conn.execute(
    "SELECT manifest_json FROM skill_versions WHERE name = ? AND version = ?",
    (sv_name, sv_version),
).fetchone()
```

`skill_versions` has `UNIQUE(name, version)`. Two workers with identical `name` and `version` in their `worker.yml` share the same row (the `ON CONFLICT(name, version) DO UPDATE` upsert means the last writer owns it). When the discover cycle runs for either worker, the `SELECT` returns whichever row exists for that `(name, version)` pair — which may be the other worker's row. The wrong `_files` dict is then carried forward, and on a container redeploy the wrong worker's source files are re-materialized to disk.

`skill_version_id` is already computed above this block as `f"sv_{worker_id}_{version}"` — it is worker-scoped and globally unique.

**Fix:** Replace the lookup:
```python
existing_sv = conn.execute(
    "SELECT manifest_json FROM skill_versions WHERE id = ?",
    (skill_version_id,),
).fetchone()
```

#### 5. `_embed_files_in_skill_version` is missing from four write paths — files never embedded, no redeploy resilience
**File:** `apps/api/main.py`

The function is called after `_register_worker_from_files`, `persist_worker_run_py`, and `update_worker_files`. It is **not** called from:

| Path | Endpoint | Impact |
|---|---|---|
| `draft_and_create_worker` Path B (LLM-generated) | `POST /workers/draft-and-create` | Most common worker creation path — LLM workers never get `_files` embedded |
| `create_worker` | `POST /workers` | REST/MCP-created workers miss embedding |
| `create_worker_from_bundle` | `POST /workers/from-bundle` | Bundle imports miss embedding |
| `rollback_worker_to_commit` | `POST /workers/{id}/rollback` | After rollback, `_files` in DB still reflect pre-rollback code; a redeploy silently undoes the rollback |

The LLM path (Path B) is the highest-traffic path — every worker created via Emily or the worker-author agent is unprotected.

**Fix:** Add `_embed_files_in_skill_version(worker_id, target_dir)` (wrapped in `try/except`) at the end of each missing path, mirroring the existing call sites.

---

### 🟡 MEDIUM

#### 6. Partial re-materialization: existing-but-empty directory bypasses the recovery check
**File:** `apps/api/run_service.py` lines 3294–3299 · `apps/api/main.py` `rematerialize_worker_from_db`

`rematerialize_worker_from_db` calls `worker_dir.mkdir(parents=True, exist_ok=True)` before writing any files. If the function is interrupted mid-loop (disk full, SIGKILL, unhandled exception), the directory exists but is partially or fully empty. The next run attempt checks `not (WORKERS_DIR / worker_id).is_dir()` — the dir exists, so re-materialization is skipped entirely. The run proceeds into the E2B sandbox with a broken worker directory and fails with a confusing deep error rather than a clean recovery attempt.

**Fix:** Write to a `.tmp` staging directory first, then `os.replace` it into position atomically. Or at minimum, also check that the entry-point file exists:
```python
worker_dir = WORKERS_DIR / worker_id
entry = worker_dir / "run.py"  # or derive from config
if not worker_dir.is_dir() or not entry.exists():
    _main.rematerialize_worker_from_db(worker_id)
```

#### 7. `_files` embedded into `manifest_json` bloats `_recipe_cache` with file contents on every request
**File:** `apps/api/db/sqlite.py` — `_recipe_cache` ContextVar

`list()` stores the full parsed `manifest_json` dict (now including `_files`) into `_recipe_cache` for every worker. On a `GET /workers` request with 20 workers each having 50 KB of embedded files, that's ~1 MB of file content held in the ContextVar for the duration of the request — content that no `get_recipe()` consumer ever reads (they access `config`, `grants`, `enabled`).

**Fix:** Strip `_files` before caching:
```python
cache[wid] = {
    "config": config,
    "grants": ...,
    ...
    # deliberately omit _files — consumers never need it via get_recipe()
}
```
Or store `_files` in a dedicated `files_json` column on `skill_versions` rather than inside `manifest_json`.

#### 8. `_build_owned_tracked_ids` fail-open exposes `worker-author` and `workspace-agent` on DB error
**File:** `apps/api/main.py` — `_build_owned_tracked_ids`

```python
except Exception:
    return frozenset(tracked)  # Fail open: treat all as owned
```

`_tracked_worker_ids()` includes `worker-author` and `workspace-agent`. When the DB query fails, the fallback returns all tracked IDs, making `_worker_hidden_from_api` return `False` (visible) for both. Neither is in `PUBLIC_STOCK_WORKER_IDS` and neither passes the `_INTERNAL_WORKER_ID_PREFIXES` check — so they appear in the user-facing `/workers` listing on any transient SQLite error.

**Fix:** Fail closed instead: `return frozenset()` — zero owned tracked IDs means all tracked workers go through the safe `get_owner()` individual fallback, which also fails open per-item but at least the list-loop path is protected.

#### 9. N+1 DB calls survive in run-list loops — only the worker-list loop was fixed
**File:** `apps/api/main.py`

`_build_owned_tracked_ids()` is correctly threaded through the `_list_visible_workers` loop (the worker-list path). But three other call sites of `_worker_hidden_from_api` were not updated:

- `_worker_source_visible_to_api()` — called per worker in source/brain-pack listings
- `_get_visible_worker()` — called on single-worker fetches (acceptable)
- `_run_visible_to_api()` — called **per run** in the runs list loop

The runs list (`GET /runs`) can return hundreds of rows, each triggering one `get_owner()` SELECT. Pre-computing a `_build_owned_tracked_ids()` frozenset once before the loop and passing it down would fix this with the same pattern as the worker-list fix.

---

### 🟢 LOW / CLEANUP

#### 10. `"*.pyc"` entry in `_WORKER_FILE_IGNORE` is dead code — glob never matches membership test
**File:** `apps/api/main.py` — `_WORKER_FILE_IGNORE`

`_should_ignore_worker_file` tests `part in _WORKER_FILE_IGNORE` where `part` is a path component string like `"foo.pyc"`. The string `"foo.pyc"` never equals the glob `"*.pyc"`. The `.pyc` check works because of the separate `part.endswith(".pyc")` guard. The frozenset entry is a no-op and misleads future maintainers into thinking glob patterns work here.

#### 11. `approve_agent_tool_approval` / `reject_agent_tool_approval` (destructive-delete variants) still omit `approval_id`
**File:** `apps/api/main.py` — `approve_action` and `reject_action` endpoints

Both `approve_destructive_action` and `reject_destructive_action` resolve the approval by `approval_id` path param, but then call `repos.approvals.approve(run_id=approval["run_id"], ...)` without forwarding `approval_id`. Now that `approve()` and `reject()` accept `approval_id`, these two endpoints should also pass it for correctness parity.

#### 12. `ApprovalsRepository` Protocol in `interface.py` not updated for `approval_id` parameter
**File:** `apps/api/db/interface.py`

The `approve()` and `reject()` Protocol stubs still don't declare `approval_id: str | None = None`. The SQLite implementation accepts it, but mypy / IDE type-checking will flag calls passing `approval_id` as an unexpected keyword argument.

---

*Analysis: 7 independent finder angles × up to 8 candidates each → dedup → verified against live source files. SyntaxError fix (`*(x,) if cond else ()` → `*((x,) if cond else ())`) is confirmed correct in both instances.*

itachi-hue:
## Medium findings addressed (follow-up to code review)

Two medium-severity findings from the review are now fixed in commit c013e34:

**Medium #2 — `_files` content bloats `_recipe_cache`**

`list()` and `get_recipe()` in `SqliteWorkerRepository` were storing the full `manifest_json` (including embedded file content) in the ContextVar recipe cache on every request. Worker files can be large — that's unnecessary memory pressure since callers only need config/grants/enabled, not blobs.

Fix: strip `_files` before writing to the cache in both places (`{k: v for k, v in mj.items() if k != "_files"}`). The raw `manifest_json` (with `_files`) is still available via a direct DB fetch when re-materialization needs it.

**Medium #4 — N+1 `_run_visible_to_api` in `system_overview`**

The `/system/overview` endpoint called `_run_visible_to_api()` once per row in three separate run-list filters. Each call hit the DB via `_get_db_worker()` — so a workspace with 100 runs in the window triggered 100+ SELECTs just to render the overview page.

Fix: `workers` (the operator-visible worker list) is already fetched earlier in the same function. Added `_visible_worker_ids = {w["id"] for w in workers}` once after that, then replaced all three `_run_visible_to_api()` calls with `row.get("worker_id") in _visible_worker_ids` — O(1) set lookups with zero extra queries.

itachi-hue:
Merged manually into main via commit 9f215e1 — squash merge with conflict resolution preserving cron_timezone, worker create locks, and git ops lock from other main commits.

PR #894, opened 2026-06-11
fix(overview): App UI v4 wireframe conformance — tiles navigate + Coming-up "See all"

itachi-hue:
## What

Brings the **Overview** page into line with the App UI v4 wireframe (`docs/design/final.html`, overview block) + spec §4, which the page had drifted from (it predates the v4 push — last touched by pre-v4 PRs).

- **Metric tiles now navigate** (spec §4: "Tiles navigate"): Runs completed/today → `/runs`, Workers active → `/workers`, Coming up → `/runs`. `MetricCard` gains an optional `href` and renders as a navigating `Link` (with hover affordance), else a plain `div` — type-safe explicit ternary, no dynamic-component spread.
- **Coming-up panel** gains its "See all" → `/runs` link, matching the wireframe and the Worker-activity panel.

The per-day hover tooltip the spec calls for is already provided by the shared `Sparkline` component's built-in per-bucket tooltip.

## Scope note

This is the one clearly-divergent page from a line-by-line read of `final.html`. The other pages (Workers, Runs, Brain, Connections, Approvals, Settings) were found to be **largely structurally conformant already** (titles, subtitles, column sets, detail tab-sets match). Remaining minor items (Runs Status/Started column order, Brain's extra "Access" column) are tracked but not in this PR. Full visual/pixel conformance (§7 DoD: light/dark × admin/member) remains unverified — it needs a render pass on a non-OOM box.

## Verification

`esbuild` transform parse OK on the edited component. Full `tsc`/visual build OOMs on the 4GB dev box; CI is authoritative for typecheck.


federicodeponte:
Closed by #2141. Verified after adversarial review: login footer links and trust content are present and build cleanly.

PR #1087, opened 2026-06-14
test: fix 33 stale tests broken by #1073 refactor → green CI

itachi-hue:
Repairs the 33 tests left stale by the #1073 oss-prep refactor (symbols moved out of main.py; patch targets / source-inspection / fake-repo interfaces went stale). **Test-only — no product source changed, no product bugs found.** Verified locally: all 11 previously-failing files pass together (148 tests), including the ordering-sensitive g1 ones. This should take main CI green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #556, opened 2026-06-07
Setup tasks: missing connection/secret/input → actionable task (Connections/Secrets tabs + global tasks)


federicodeponte:
## Setup tasks: missing connection/secret/input becomes an actionable task

Federico (2026-06-07): when a worker is **created or imported** that needs a connection/secret/input the workspace doesn't have, it should become an **actionable task to complete that auth flow / add that secret** — surfaced in 3 places:
1. **Inline on the worker** — it stays in a "Needs setup" state and can't run until complete.
2. **Under the Connections / Secrets tabs** — "Connect Gmail (required by Granola→HubSpot)" / "Add `APIFY_API_KEY`".
3. **In the global tasks / needs-attention list.**

This is the **user-facing complement to #551** (block activation until required connections/secrets satisfied) — instead of a silent failure, it's a clear to-do.

### Extensions (other intuitive flows, same spirit)
- **Bulk import** → one consolidated setup checklist (all connections/secrets needed across the imported workers), each item a task.
- **First-run onboarding** = "3 things to get your first worker running: connect X, add Y" as tasks (activation lever).
- **Global needs-attention inbox** aggregates: pending approvals + failing workers + incomplete setup tasks — one daily surface.
- A worker's run page, when blocked, shows "Blocked: missing `APIFY_API_KEY` → [Add it]" deep-linking to the secret form.
"Be super intuitive about everything." Design with on-brand variants first (per the UI rule) before building.

itachi-hue:
Needs backend + design: surfacing missing connections/secrets as actionable tasks requires a new task data model (or piggybacking on an existing one), integration with worker activation (#551), and a global needs-attention inbox. Too broad for a single code fix. Skipping this batch pending a design pass.

itachi-hue:
Skipping this batch. Setup tasks (missing connection/secret/input -> actionable task with Connections/Secrets tabs + global tasks view) require a new task-tracking data model on the backend and a tasks UI surface. Too large for this batch without a design spec.

itachi-hue:
Fixed in PR #576 (commit 0251323, branch fix/issues-batch6-2026-06-07).

**Backend:**
- WorkerSummary and WorkerDetail now include missing_secrets: List[str] and missing_connections: List[str] — the specific items required but not yet configured
- New _available_connection_slugs_for_user() helper queries active connection rows
- Both worker list and detail endpoints compute and populate these fields

**Frontend:**
[...]

PR #1095, opened 2026-06-14
test(web): fix stale #556 overview grep-tests (point at routers/overview.py)

federicodeponte:
## What
Two `#556 Surface 3` grep-tests in `apps/web/tests/fl-batch-10.test.ts` still read `apps/api/main.py`, but the **#1073 refactor moved `system_overview` into `apps/api/routers/overview.py`**. Result: 0 symbol matches in main.py → Web tests FAIL on every PR off main (incl. #1079, #1087).

## Evidence
- `routers/overview.py` contains all 7 grepped symbols — 13 matches.
- `main.py` has **0**.
- Feature fully intact; **stale test path**, not a regression.

## Fix
Point the 2 #556 tests at `routers/overview.py`. Test-only, single file, +5/-2.

## Verify
`npx tsx tests/fl-batch-10.test.ts` → **8/8 passed** (was 6/8).

## Relationship to #1087
#1087 (fix 33 stale tests from #1073) does **not** touch `fl-batch-10.test.ts` — these 2 are the gap it missed. No file overlap; complementary.


federicodeponte:
Superseded by #1087 (merged `aecff96b`), which fixed the same 6 `fl-batch` files via per-file repoint plus the pytest staleness — main is green. Closing.

Optional future hardening: this PR's `tests/_apisrc.ts` `apiAll()` helper greps the whole `apps/api` tree so these existence-checks survive the *next* file move too (vs per-file pins that will re-break on the next refactor). Happy to re-open as a small standalone follow-up if we want that resilience — not urgent, tests are green either way.

PR #1193, opened 2026-06-15
fix(security): move OSS API secret from localStorage to sessionStorage

### LINKED ISSUE #1185: [HIGH] OSS API secret is persisted in browser localStorage
`apps/web/components/CliCommandPanel.tsx` reads and writes the operator API secret (`floom_secret` / `FLOOM_SECRET` / `workeros_api_secret`) to `window.localStorage` (lines 13, 28-35, 178, 233).

Any XSS payload or malicious browser extension running in the dashboard origin can exfiltrate this single-user secret and gain full API access.

Suggested fix: do not persist the secret client-side. Keep it in component state / memory only, or use `sessionStorage` with a clear warning if persistence is required. If localStorage is kept for UX, encrypt it with a key derived from a fresh session cookie or use the Web Crypto API before storing.

/cc @vivek

federicodeponte:
## Summary

- **Root cause:** `CliCommandPanel.tsx` was writing the OSS API secret to `window.localStorage` under three keys (`floom_secret`, `FLOOM_SECRET`, `workeros_api_secret`). `localStorage` is persistent across browser restarts and accessible to any JavaScript running on the same origin — a single XSS payload or malicious browser extension can silently exfiltrate it.
- **Fix:** Replace all `localStorage` read/write/clear calls with `sessionStorage`. `sessionStorage` is scoped to a single tab and cleared when the tab closes. Added a `beforeunload` handler that explicitly removes the key on navigation/close so the value does not linger in session-restore or memory-swapped tab state.
- **Legacy eviction:** Added `evictLegacyLocalStorage()` called on component mount that removes the three legacy `localStorage` keys for any existing user whose browser already has a stored secret from a previous build.
- **Auth flow unaffected:** The actual authentication gate is the `workeros_session` httpOnly cookie (verified in `middleware.ts` and derived/verified in `lib/web-session.ts`). The secret in sessionStorage is only a UX cache for the token-display panel — it is never sent directly to the API by the web app.

## Before / After

| | Before | After |
|---|---|---|
| Storage mechanism | `window.localStorage` | `window.sessionStorage` |
| Persists across browser restart | Yes | No |
| Accessible to other tabs | Yes | No |
| Cleared on tab close | No | Yes (+ beforeunload) |
| Old keys evicted on upgrade | No | Yes (mount-time cleanup) |
| `localStorage.getItem("floom_secret")` readable | YES | **empty** |

## Test plan

- [x] All 440 existing tests pass (`npx vitest run` — 85 test files, 440 tests)
- [x] TypeScript compiles clean on changed file (pre-existing errors in test files only)
- [ ] Manual: generate a token in Settings → API access tab; verify `sessionStorage.getItem("workeros_oss_secret")` contains the token and `localStorage.getItem("floom_secret")` is `null`
- [ ] Manual: close the tab and reopen — verify the token is no longer pre-filled (sessionStorage cleared)
- [ ] Manual: verify the auth flow (login → dashboard) still works end-to-end

Closes #1185


federicodeponte:
Superseded by current main. The production fix for #1185 is already present in e79eabc6 / #1200: CliCommandPanel no longer writes the OSS API secret to localStorage, uses sessionStorage, and removes legacy localStorage keys. Closing without merging to avoid replaying a stale branch over current main.

Issue #1182, opened 2026-06-15
[HIGH] /api/proxy follows backend redirects server-side for non-OAuth requests


federicodeponte:
`apps/web/app/api/proxy/[...path]/route.ts:123` sets `redirect: isConnectionCallback ? "manual" : "follow"`. For every non-OAuth request, the Next.js proxy follows 3xx responses from the backend and streams the final response back to the browser.

If the backend ever returns a `Location` to an external host, the proxy will fetch it server-side, creating both an open-redirect vector and an SSRF surface (the proxy carries the user's JWT/cookie).

`safeProxyLocation` only runs when `redirect: "manual"`. Either switch all proxy requests to `"manual"` and consistently sanitize/forward only safe `Location` headers, or validate the final response URL before streaming it back.

/cc @vivek

federicodeponte:
✅ Fixed and **LIVE on prod**. Merged to engine main (`815cafdd`) via the Tier-A cherry (PR #1200) / cloud security (PR #313), deployed to Railway + Vercel (cloud main `287c4da1`), smoke gate passed all 11 routes, backend tests green. Verified prod serves the new engine; em-dashes confirmed 0 on the live landing. Closing.

PR #1194, opened 2026-06-15
fix(security): block SSRF in /connections/test + /api/proxy redirects

### LINKED ISSUE #1180: [HIGH] SSRF via POST /connections/{id}/test — MCP URL not re-validated at dial time
`POST /connections/{id}/test` in `apps/api/routers/connections.py` (around lines 1461 and 1488) probes the stored `mcp_url` with `httpx.post(...)` to perform a live connection test.

The URL is validated once at creation, but it is **not** re-validated at test time. An attacker who can change the stored URL (DNS rebinding, a compromised DB/admin, or a race after creation) can make the API host probe internal/metadata endpoints such as `169.254.169.254`, `localhost`, or RFC1918 addresses.

Suggested fix: call `assert_safe_outbound_mcp_url(probe_url)` immediately before each `httpx.post`/`httpx.get` in the test path and fail closed with `ConnectionTestResult(status="failed")` if the URL is unsafe.

/cc @vivek

federicodeponte:
Fixes #1180 and #1182.

### #1180 — DNS-rebinding SSRF in `POST /connections/{id}/test`

`test_connection` reads `mcp_url` from the database and immediately calls `httpx.post(probe_url)` without re-running the SSRF guard. The URL is validated once at creation time (via `assert_safe_outbound_mcp_url`), but a DNS rebind between creation and test time can point a public-looking hostname at `169.254.169.254`, `10.x.x.x`, `127.0.0.1`, etc., causing the API server to probe internal endpoints carrying the user's auth headers.

### #1182 — Server-side redirect following in `/api/proxy`

`redirect: "follow"` was used for every non-OAuth proxy request. When the backend returns a `3xx Location` pointing at an internal host, Node's `fetch()` silently follows it server-side — `safeProxyLocation()` is never reached because the redirect is consumed before the response is inspected.

## Fix

**`apps/api/routers/connections.py`**: call `assert_safe_outbound_mcp_url(mcp_url)` immediately before the first `httpx.post`, fail closed with `status="failed"` + a clear SSRF-guard reason. The guard re-resolves the hostname on every call, defeating DNS rebinding.

**`apps/web/app/api/proxy/[...path]/route.ts`**: switch all requests to `redirect: "manual"` unconditionally. The existing `safeProxyLocation()` function already strips Location values not relative to the app or API origin, so every redirect — OAuth and non-OAuth — is now sanitised before being forwarded to the browser.

## Exploit closed — before / after evidence

| Scenario | Before fix | After fix |
|---|---|---|
| Connection test, URL now resolves to `169.254.169.254` (DNS rebind) | `httpx.post` dials metadata endpoint | `assert_safe_outbound_mcp_url` raises, returns `{"status":"failed","reason":"Connection URL rejected by SSRF guard: ..."}` |
| DB-tampered `mcp_url = http://169.254.169.254/…` | Same as above | Same block — guard runs on the literal IP |
| Proxy receives `302 Location: http://169.254.169.254/…` | `fetch` follows redirect server-side, streams internal response to client | `redirect:"manual"` → 302 reaches our code, `safeProxyLocation` returns `null`, Location header is stripped, browser receives 302 with no Location |

## Regression tests

- `test_connection_test_endpoint_rejects_internal_url` — DNS-rebinding simulation
- `test_connection_test_endpoint_rejects_metadata_literal` — DB-tamper simulation
- `apps/web/tests/proxy-ssrf-1182.test.ts` — 5 cases: `redirect:"manual"` on GET + POST, internal Location stripped, external Location stripped, safe relative Location preserved

All 28 existing Python SSRF tests + 13 proxy TS tests pass.


federicodeponte:
Superseded by current main. The production SSRF protections from #1180/#1182 are already present in e79eabc6 / #1200: /connections/{id}/test re-validates the stored MCP URL at dial time, and /api/proxy uses redirect:"manual" before sanitizing Location. Closing without merging because this PR is stale and its API check is red.

PR #1490, opened 2026-06-18
fix(cloud): bound runs list hot path

federicodeponte:
## Summary
- add a Supabase `list_operator_visible` fast path for `/runs` that uses a lightweight select, no exact count, cheap has-more semantics, and keyset ordering
- remove `output_json` from run list selects; detail reads still fetch full run payloads
- configure a larger Supabase httpx connection pool and retry idempotent GET/HEAD/OPTIONS requests on dropped pooled connections
- add concurrent keyset indexes for `public.runs` list ordering and status-filtered list ordering

## Why
Live Railway logs showed `/runs` list taking about 40s for 191 rows and Supabase `RemoteProtocolError` under load. The slow path selected large payload columns and requested exact PostgREST counts. This PR keeps writes single-shot and only hardens read paths.

## Verification
- `python3 -m py_compile apps/api/config.py apps/api/db/supabase_repos.py tests/test_supabase_repos.py`
- `python3 -m ruff check apps/api/config.py tests/test_supabase_repos.py`
- `python3 -m pytest tests/test_supabase_repos.py::test_run_get_falls_back_to_owned_exact_id_when_workspace_cookie_is_stale tests/test_supabase_repos.py::test_run_list_operator_visible_uses_lightweight_select_and_cheap_total`

Full `apps/api/db/supabase_repos.py` lint still has existing E402/F811/F401 violations unrelated to this patch.

## Migration
`supabase/migrations/0034_runs_list_keyset_indexes.sql` uses `CREATE INDEX CONCURRENTLY`. Apply it before the API deploy on the hot runs table.

## Deploy
Do not deploy from this PR. The human will deploy after current benchmarks finish.


itachi-hue:
Closing because cloud release/deployment branches should not live in the WorkerOS repo. The branch history is preserved in the private branch archive.

Issue #1558, opened 2026-06-19
[bug] False 'Item not found / old ID format' toast on valid deep-links (lists are cache-first)


federicodeponte:
**The message is misleading — there is NO id-format validation anywhere** (grepped: no prefix/uuid/slug check, no cloud-vs-engine transform). It's a **false 'not found'**.

**Root cause:** `components/collection/CollectionView.tsx:54-57` sets `selMissing` whenever the `?sel=<id>` param isn't in the *currently-loaded* `config.items` → `Collection.tsx:60` fires the toast. But the lists are **partial/cache-first**:
- **Runs:** `RunsCollection.tsx:188` PAGE_SIZE=50 — loads only the 50 most recent; deep-linking/sharing/bookmarking an older run → false toast even though `api.runs.get(id)` would return it. **Most likely surface Federico hit.**
- **Workers:** cache-first (30s) + filters system workers; a just-created worker / deep link → same false toast.

The cloud overlay already documents + partially fixes this **for workers only** (`web/app/workers/WorkersCollection.tsx` ~193-205 `detailToSummary` + ~1785-1830 selParam hydration). The engine workers list has ZERO hydration; **runs has it nowhere.**

**Fix:** generalize into shared `CollectionView`/`Collection`: `config.resolveMissing?(id) => Promise<T|null>`; on `selMissing` call it (`api.runs.get` / `api.workers.get` / …), merge into items, only toast if it genuinely returns null. **Also fix the copy** — drop 'old ID format' (never the cause).

itachi-hue:
Verified fixed in current web code: CollectionView now supports `resolveMissing(id)` and Workers/Runs hydrate deep-linked ids by direct fetch before showing any not-found toast. Focused regression tests passed: `tests/collection-view.dom.test.tsx` and the #1877 follow-up tests.

PR #1560, opened 2026-06-19
fix: false 'not found' toast on valid deep-links (#1558)

federicodeponte:
## Problem

Deep-linking a valid `?sel=<id>` that isn't in the currently-loaded list false-toasted **"Item not found; it may have been deleted or the link uses an old ID format."** There is **no actual id-format problem** — the lists are partial:

- **Runs** is paginated (`RunsCollection` `PAGE_SIZE=50`), so a run beyond the first page isn't in `config.items`.
- **Workers** is cache-first (staleTime 30s) and filters out system workers, so a just-created or filtered worker isn't in the loaded list.

`CollectionView` set `selMissing` whenever `?sel` wasn't in `config.items`, and `Collection` fired the toast. The id was fine; the list was stale/partial.

## Fix (generalizes the prior workers-only hydration into the shared grammar)

1. **`config.resolveMissing?(id): Promise<T | null>`** added to the Collection grammar. When `?sel` is absent from the loaded list, `CollectionView` calls it **once per id** (guarded by a ref against loops/duplicate fetches), **merges** a non-null result into the displayed items so the detail opens, and **does not toast**. Only a `null` result or a throw is a genuine miss → `onInvalidSel` (toast + rest). Collections **without** `resolveMissing` keep the existing toast-on-miss behavior (no regression).
2. **Wired** `resolveMissing` for the deep-linking collections:
   - **Runs**: `api.runs.get(id)` projected to `RunSummary` (+ warms the detail-pane cache).
   - **Workers**: `api.workers.get(id)` via the existing `detailToSummary`. The prior bespoke per-component `selParam` hydration effect is **removed** — the generic mechanism replaces it (DRY), and CollectionView's merged set persists the hydrated item across list refetches on its own.
   - **Connections** intentionally left as-is: its list is a synthesized union with no single by-id resolver, so it keeps the (correct) toast-on-genuine-miss behavior.
3. **Copy fixed** in `Collection.tsx`: dropped the misleading "or the link uses an old ID format"; the toast now reads **"Couldn't open that item; it may have been deleted."** and only shows when `resolveMissing` truly fails (or the collection has none).

## Verification

- `npx tsc --noEmit` — 0 errors
- `npx next build` — 0 errors
- `npm run lint` — 0 errors (pre-existing warnings only)
- `npx vitest run` — 573 passed; only the pre-existing unrelated `login-secrets-render-987-988` failure (`api.me is not a function`) remains
- New tests: a `?sel` id absent from the list but resolvable via `resolveMissing` opens the detail **without** a toast; a genuine miss (`null` / throw) **still toasts** with the new copy; `resolveMissing` fires at most once per id.

Closes #1558


itachi-hue:

Issue #1561, opened 2026-06-19
[bug] Light-mode render failures — Run Output + Settings Members stay blank/skeleton (dark works)


federicodeponte:
Theme-dependent rendering/data bug found in the audit:
- **Run detail → Output tab (light mode):** 5 skeleton bars that NEVER resolve (waited 4s+). In DARK mode the same run loads real output at ~5s. Light mode consistently stays on skeletons. (`runs-detail-loaded-light-desktop.png` vs `runs-detail-dark-desktop-5s.png`). Mobile run-output also skeleton-only.
- **Settings → Members panel (light mode):** ~60% blank white void; DARK mode shows the real member list (fede@/george@/vivek@). (`settings-members-detail-light-desktop.png` vs `settings-members-dark-desktop.png`).
Both render fine in dark, fail in light → a theme-conditional load/render path. Likely a shared cause. apps/web. UI.

federicodeponte:
**Correction: NOT theme-related.** Root-cause: an intermittent async-fetch race in `useRunDetail` (`app/runs/RunsCollection.tsx:52-76`) and the parallel `MembersSettingsPanel` (`app/settings/page.tsx:2198`). Both fetch in a useEffect with an `alive` cancel flag; when cleanup runs before the promise resolves (rapid select / remount / cache transition) the resolved `setState` is DROPPED → stuck on skeleton until an unrelated re-render (e.g. the theme toggle) papers over it. That's why 'dark works' — toggling theme just re-renders. **Fix:** write to the detail cache even when `!alive`, and re-derive state from cache synchronously on remount, so a dropped fetch doesn't strand the skeleton. Same fix for both. Retitle: 'intermittent skeleton-forever on Run Output + Members (async-fetch race)'.

Issue #1544, opened 2026-06-19
[mobile] Redesign not mobile-audited — P0: home blank on mobile + header-squish across pages


federicodeponte:
Audited the redesign at 390px (iPhone) on app-preview. Desktop is fine; **mobile was never tested** and has real issues. Screenshots on AX41 `/tmp/mobile-audit/`.

**P0 — mobile home is unusable.** `/app` (and `/app/?create=1`) at 390px render ONLY the Emily radar mark — no greeting, no pills, no composer. Emily's fullscreen empty state doesn't render its content on mobile (works perfectly on desktop). Likely the mobile path (EmilyMobileSheet / the centered-composer empty state) isn't wired for mobile. **Verify on a real device + check if it's on prod too** (prod has the redesign without the round-3 centered-composer change, so confirm whether mobile-home was already broken pre-U1 or U1 regressed it).

**P1 — page-header squish (connections / workers / settings).** At 390px the header description ('Apps, MCP servers and secrets…', 'Your AI workers.', 'Workspace settings…, kept separate.') squeezes into a narrow multi-line column beside the inline stat chips instead of stacking. On /connections the rightmost stat ('0 re…') is clipped at the viewport edge. Fix: stack description above stats / wrap stats below at small widths.

**P2 — polish.** Mobile sidebar drawer shows the '⌘K' shortcut hint (meaningless on phones). Landing has ~1-2px horizontal overflow (faint scrollbar). Landing hero composer placeholder not visible.

These are apps/web (engine). Owner: UI. Tag whoever picks up mobile.

federicodeponte:
Full mobile audit confirms + adds: **P0** home blank on mobile (light+dark, stuck spinner). **P1** header-squish on workers/settings/connections (description wraps beside inline stats; connections '0 re…' clipped). **P1** run detail on mobile = no run-list + skeleton-only output = unusable. **P2** runs table columns crushed (WORKER→'N…', STARTED vertical). **P2** primary 'New worker' CTA hidden behind the hamburger on mobile. Screenshots /tmp/full-audit/.

itachi-hue:
Verified fixed on current main for the reported mobile shell regression. 	ests/tablet-shell-breakpoint-1544.test.ts passed in the targeted web run, and the code now uses the mobile/tablet shell breakpoint with the mobile Emily affordance.

PR #1568, opened 2026-06-19
fix(render): mobile home shows Emily+workers not blank radar (#1544); harden run/members fetch race (#1561)

federicodeponte:
### FIX 1 — P0-1 mobile home blank (from #1544)
On mobile, the home pane rendered the bare `aria-hidden` `EmilyRadarMark` because `emilyFull` (`fullscreen && isDesktop`) is always `false` on mobile, so the page pane never hid and `HomePane` fell through to the radar placeholder.

- `HomePane` now renders `WorkersCollection` (the same component `/workers` uses) whenever `!isDesktop` — never the radar.
- `EmilyMobileSheet` already auto-opens on the home route, so the user lands in Emily (greeting + "Message Emily…" composer); closing it reveals the Workers list — never a blank radar.
- The radar placeholder is now strictly desktop-only (the brief pre-fullscreen frame). Desktop behavior is unchanged.
- Extracted the shared `useIsDesktop` hook into `lib/use-is-desktop.ts` (was a private copy in `AppShell`); `AppShell` now imports it.

### FIX 2 — P0-2 / P1-5 async-fetch race (closes #1561)
`useRunDetail` (`app/runs/RunsCollection.tsx`) and `MembersSettingsPanel` (`app/settings/page.tsx`) could strand the skeleton when a fetch resolved after the consuming component unmounted — the `alive`/cancel flag dropped the resolving `setState`.

Both hooks now:
- Write the resolved result into the module/detail **cache even when `!alive`**, so a remount can read it synchronously.
- Derive **initial state synchronously from the cache** on mount, so a remount shows data immediately without waiting on a new fetch.
- No refetch loops introduced.

Added `tests/run-detail-cache-on-cancel.dom.test.tsx` proving a post-unmount fetch still populates the cache and the next mount reads it synchronously.

## Verification
- `npx tsc --noEmit` — 0 errors
- `npx next build` — success
- `npm run lint` — 0 errors (pre-existing warnings only)
- `npx vitest run` — 570 pass, 1 pre-existing failure (`login-secrets` `api.me is not a function`, unrelated)

closes #1561; fixes mobile-home P0 from #1544


itachi-hue:

Issue #1555, opened 2026-06-19
[HIGH] Worker run output_json persisted without scrub_secrets


federicodeponte:
## Summary
Malicious run.py can read secrets.json / .env.local and put keys in result.outputs. Caller retrieves via GET /runs/{id} output_json. Logs are scrubbed; outputs are not.

## Evidence
- `apps/api/runner_sandbox/e2b_driver.py:944-945`
- `apps/api/run_service.py:3285-3288,1980-1984` — persist without scrub
- Log scrub at run_service.py:3140 does not apply to outputs

## Fix
Apply scrub_secrets to output_json before persist and on GET response.

itachi-hue:
RCA + fix are in review: https://github.com/floomhq/workeros/pull/1572

RCA: run finalization persisted output_json directly, including the approval-pending output path, and read/export/download paths returned stored output as-is.

Fix in PR:
- recursively scrub declared worker secret values before persisting run output
- apply the same scrub to the approval-pending path that bypassed normal finalization
- add read-side defense for run detail, single-run ZIP download, and bulk export so legacy rows are also redacted

Tests:
- 	est_run_service_scrubs_secret_values_before_persisting_output
- 	est_run_detail_redacts_legacy_persisted_output_secrets
- included in the 103-test backend security regression set documented on the PR

itachi-hue:
Fixed by merged PR https://github.com/floomhq/workeros/pull/1572 (merge commit 0e83d63). RCA, fix details, and test coverage are documented above in this issue.

Issue #1554, opened 2026-06-19
[HIGH] E2B worker-call token always issued at depth=0 — unbounded nested sandbox spawning


federicodeponte:
## Summary
When worker has `calls: [self]` or cyclic call graph, E2B driver always issues worker-call token with `depth=0`. Server gate at main.py always sees depth=0; client-side check reads static env.

## Evidence
- `apps/api/runner_sandbox/e2b_driver.py:682-686` — `depth=0` hardcoded
- `apps/api/main.py:10467-10468` — depth check always sees 0

## Exploit
Worker with self-call or cycle spawns unbounded E2B VMs, exhausting quota/cost.

## Fix
Track and increment call depth from parent run context; enforce max depth server-side.

itachi-hue:
﻿Verified under the same review PR: https://github.com/floomhq/workeros/pull/1572

This was already addressed on current main before this branch: worker-call depth is parsed/propagated and nested E2B calls are capped instead of always minting depth=0 tokens.

Regression coverage included in the PR verification:
- apps/api/tests/test_994_call_depth_cap.py
- apps/api/tests/test_worker_calls.py

itachi-hue:
Fixed by merged PR https://github.com/floomhq/workeros/pull/1572 (merge commit 0e83d63). RCA, fix details, and test coverage are documented above in this issue.

Issue #1553, opened 2026-06-19
[HIGH] Standalone share-link revoke DELETE is a no-op (routes never implemented)


federicodeponte:
## Summary
Frontend calls DELETE on share links; backend has no DELETE routes for standalone/worker share links. `_revoke_standalone_share_link` only wired for review packs.

## Evidence
- `apps/web/lib/api.ts:311-314,452-454,714-717` — DELETE calls
- `apps/api/main.py:6874-6931` — no DELETE handler for standalone shares
- `routers/review_pack.py:91-100` — revoke only for review packs

## Impact
Owner toggles "revoke" → 404/405 → link stays live forever.

## Fix
Implement DELETE revoke routes; route all share ops through `services/share_links.py` with token rotation.

itachi-hue:
﻿Verified under the same review PR: https://github.com/floomhq/workeros/pull/1572

This was already addressed on current main before this branch: standalone share-link DELETE routes revoke links instead of being no-ops.

Regression coverage included in the PR verification:
- apps/api/tests/test_share_link_revoke.py

itachi-hue:
Fixed by merged PR https://github.com/floomhq/workeros/pull/1572 (merge commit 0e83d63). RCA, fix details, and test coverage are documented above in this issue.

Issue #1552, opened 2026-06-19
[HIGH] Composio OAuth callback has no state parameter (account-linking CSRF)


federicodeponte:
## Summary
Composio connection OAuth flow has no `state`/nonce binding callback to initiating session. Slack OAuth uses state HMAC; Composio does not.

## Evidence
- `apps/api/composio_client.py:467-490` — OAuth URL generation
- `apps/api/main.py:14096-14279` — callback updates row by `composio_connection_id` without session binding

## Exploit
1. Attacker initiates `POST /connections` → gets Composio redirect tied to attacker's user_id
2. Sends victim the Composio hosted link
3. Victim completes provider OAuth
4. Callback attaches victim's connected account to attacker's connection row

Distinct from #1465 (status query param trust).

## Fix
Add HMAC state nonce binding user_id/session; verify on callback.

itachi-hue:
﻿RCA + fix are in review: https://github.com/floomhq/workeros/pull/1572

RCA: Composio OAuth initiation did not include a state value, so callback handling could not prove the browser session that received the callback was the same user/session that initiated the connection.

Fix in PR:
- issue short-lived HMAC-signed OAuth state bound to the initiating user
- append that state to the callback URL sent to Composio
- forward state through the frontend callback proxy
- require state user == active session user == stored connection owner before remote status checks or activation
- keep the existing remote status check so ?status=success alone cannot promote an initiated connection

Tests cover signed state emission, missing state rejection, mismatched session rejection, and remote initiated not being promoted by query params.

itachi-hue:
Fixed by merged PR https://github.com/floomhq/workeros/pull/1572 (merge commit 0e83d63). RCA, fix details, and test coverage are documented above in this issue.

PR #1572, opened 2026-06-19
fix: harden OSS security boundaries

itachi-hue:
## Summary
Fixes the OSS high-priority security batch:

- #1555: scrub worker output before persistence and again on run read/export/download paths
- #1552: add signed OAuth `state` to Composio connect URLs and require state/session owner match on callback
- #1549: allowlist backend-supplied run parts SSE paths before opening `EventSource`
- #1525: treat Cloud CLI device-flow `429` polling responses as `slow_down` instead of fatal

Also verifies the already-present fixes for:

- #1554: E2B worker-call depth propagation/cap
- #1553: standalone share-link revoke DELETE
- #1551: one-time OSS magic-link nonce consumption
- #1547: Emily `runs__cancel` ownership enforcement

### #1555 output_json persisted without scrub_secrets
RCA: run finalization persisted `output_json` directly, including approval-pending output paths, and read paths returned stored output as-is.
Fix: added recursive secret-value scrubbing before `output_json` persistence, applied it to the approval-pending bypass path, and added defense-in-depth redaction for run detail, single-run download, and bulk export so legacy rows are not leaked.
Tests: `test_run_service_scrubs_secret_values_before_persisting_output`, `test_run_detail_redacts_legacy_persisted_output_secrets`.

### #1552 Composio OAuth callback missing state
RCA: OAuth initiation supplied a callback URL without an anti-CSRF state value, and callback trusted connection IDs/status without proving the browser session initiated the flow.
Fix: issue a short-lived HMAC-signed state bound to the initiating user, append it to callback URL, forward it through the web callback proxy, and require callback state user == active session user == stored connection owner before remote status checks or activation.
Tests: callback URL includes signed state; missing/mismatched state does not activate; remote `initiated` status is not promoted by `?status=success`.

### #1549 backend-supplied streams.parts EventSource
RCA: chat UI opened `EventSource` on a backend-supplied path from tool-card metadata.
Fix: allowlist only exact `/runs/<id>/stream` paths with safe run-id characters; reject external URLs, query strings, and other backend paths.
Tests: `useChatStream` allowlist test.

### #1525 login --cloud device-flow 429 polling
RCA: Cloud polling could return 429 while the user had not approved the device flow yet, and the CLI treated that as fatal instead of OAuth-style `slow_down`.
Fix: in Cloud login polling, handle 429 by sleeping at least the server interval / 5 seconds and continuing.
Tests: MCP/CLI cloud-mode source-level regression.

## Verification

- `python -m pytest apps/api/tests/db/test_sqlite_runs.py apps/api/tests/test_run_inputs_exposed.py apps/api/tests/test_g1_login_signing.py apps/api/tests/test_994_call_depth_cap.py apps/api/tests/test_worker_calls.py apps/api/tests/test_share_link_revoke.py apps/api/tests/test_emily_cross_tenant_237_238.py tests/test_connections_backend.py -q` -> 103 passed
- `npm test -- useChatStream.test.ts` in `apps/web` -> 14 passed
- `npm test -- cloud-mode.test.js` in `apps/mcp` -> 69 passed
- `python -m ruff check apps/api/routers/connections.py apps/api/routers/runs.py apps/api/run_service.py apps/api/tests/db/test_sqlite_runs.py apps/api/tests/test_run_inputs_exposed.py tests/test_connections_backend.py` -> passed
- `npx tsc --noEmit` in `apps/web` -> passed
- `git diff --check` -> passed, only Windows LF->CRLF warnings


itachi-hue:
## Review — strong security PR, approve with one cloud deploy note

4 well-targeted fixes, **each with tests**, scope clean (app code only), no telemetry, no hardcoded secrets, and no conflict with the merged `run_service.py` recovery code.

**Fix quality — verified ✓**
- **#1552 OAuth state (CSRF):** HMAC-signed, user-bound, 600s TTL, `hmac.compare_digest` (timing-safe), and the callback enforces the triple match `state_user == session_user == connection owner`, failing closed. Textbook defense.
- **#1555 output scrubbing:** recursive `_scrub_run_output` before persist + approval-pending path + run detail + single download + bulk export, with try/except defense-in-depth for legacy rows. Tested (persist + legacy redaction).
- **#1549 SSE allowlist:** tight `safeRunPartsStreamPath` with strong negative tests (rejects `/logs/stream`, `?x=/connections` query injection, and absolute `https://attacker.example/...`).
- **#1525 CLI 429→slow_down:** correct (`status === 429 && isCloud`), tested.

**Flags**
1. ⚠️ **Deploy note (cloud) — real.** The OAuth state secret resolves `WORKEROS_OAUTH_STATE_SECRET → WORKEROS_MAGIC_LINK_SECRET → FLOOM_SECRET → per-process random`. The cloud deliberately strips `FLOOM_SECRET` (#992), so a multi-replica cloud deploy would fall through to the per-process random → states signed by one replica fail on another (and after restart) → intermittent OAuth-callback failures. **Cloud must set `WORKEROS_OAUTH_STATE_SECRET` (or `WORKEROS_MAGIC_LINK_SECRET`) explicitly.** Fine for single-process OSS.
2. #1552 now requires `session_user == owner` at callback (fails closed if no session). Worth one manual OAuth-connect smoke on OSS to confirm a session cookie is always present there.
3. CI red = same inherited `main`-red (#1201) + Vercel project-access; give Secret-scan the same human glance since this is a security PR.

Nice work — solid hardening. The only thing I'd gate on is the cloud `WORKEROS_OAUTH_STATE_SECRET` env so OAuth doesn't silently break across replicas.

_Automated review (Claude Code)._

Issue #1562, opened 2026-06-19
[bug] Approvals detail — 'PROPOSED OUTPUT' section is empty (can't see what you're approving)


federicodeponte:
In the approval detail (desktop + mobile, both modes), the 'PROPOSED OUTPUT' section renders with NO content — no proposal, no placeholder. The user is asked to Approve/Reject without seeing what the worker proposes. Core HITL failure. `approvals-detail-v2-light-desktop.png`, `approvals-detail-light-mobile.png`. apps/web.

itachi-hue:
Verified fixed in current web code: ApprovalReviewBody renders proposed output from preview/decision payload and shows a fallback only when genuinely absent. Regression passed: `tests/approval-proposed-output-1562.dom.test.tsx`.

PR #1578, opened 2026-06-20
fix(approvals): render proposed output (preview) in approval detail (#1562)

federicodeponte:
## Problem (#1562)

The approval-detail **Proposed output** column rendered **empty** for the common case: a worker emits a plain `preview` string with no structured action items (`decision_input_json` is a flat object or `{}`). No preview, and not even the empty-state placeholder.

## Verdict: UI bug (data present, not rendered)

Verified against `app-preview.floom.dev` as `fede@floom.dev`:

- The frontend proxy `/app/api/proxy/approvals?status=pending` returns the pending approval **with `preview` populated**: `"draft for Test Prospect Beta"` (`preview_payload: null`, `decision_input_json` a flat object with no array).
- The live detail page rendered `PROPOSED OUTPUT` immediately followed by the footnote — the content area was empty.

So the API returns a proposed-output field and it is populated; the UI just failed to render it.

## Root cause

In `ProposedOutput` (`ApprovalReviewBody.tsx`):

```tsx
const items = <ApprovalActionItems decisionInput={di} />;
if (items) return <div className="c-appr-proposed">{items}</div>;
```

`items` is a **JSX element**, which is always truthy — even when `ApprovalActionItems` renders `null` (no array in `decision_input_json`). So the branch was always taken, emitting an **empty div**, and the `approval.preview` branch + the "No proposed output attached" fallback below were **dead code**.

## Fix

- Export `hasActionItems(decisionInput)` from `ApprovalActionItems.tsx` (single source of truth, reuses the existing `extractItems`).
- Branch on `hasActionItems(di)` instead of the JSX element. Render order is now: structured items when present → else the `preview` string via `GenericOutput` → else the empty-state placeholder, and only when genuinely empty.
- This also fixes the shared `/approvals/review` page, which falls through to the same `ProposedOutput` when there is no file-preview override.

## Tests / gates

- New `tests/approval-proposed-output-1562.dom.test.tsx` locks all three paths (preview-only regression, structured items, genuinely empty).
- `tsc --noEmit`: 0 errors. `npm run lint`: 0 errors. `next build`: OK.
- `vitest`: 572/573 pass. The 1 failure is the pre-existing `login-secrets-render-987-988` env failure (unrelated to this change).


itachi-hue:

Issue #1565, opened 2026-06-19
[polish] UI audit P2 batch (~16 items)


federicodeponte:
From the full audit (screenshots /tmp/full-audit/): workers all show 'Draft' badge despite '5 active' (P2-1); success-rate '29%' shown unlabeled (P2-2); worker-detail OUTPUTS box + description truncated (P2-3/P2-11); run-detail sidebar entries all identical 'NovaSearch V5 / Completed 14h ago' — no disambiguation (P2-16); dark-mode faint/near-invisible borders on approvals textarea + status dots (P2-7/P2-17); settings sidebar over-truncates ('Personal access tok…') (P2-8); Runs/Approvals tab badges render inline not as superscript (P2-18); Composio generic logo (P2-19); MCP popup small mono + border-radius mismatch (P2-10); **server internals exposed in Settings General — Version/Python/Runner:e2b/Started-at** (P2-20, confirm intended); small unreadable mono in run-output Reason col (P2-9). apps/web.

itachi-hue:
Closing as not planned for the current launch sweep. This is stale, non-blocking, or product-backlog work; we are keeping the open issue list focused on must-fix and medium-priority launch work.

itachi-hue:
Reopening after interrupted stale-cleanup batch: product/design/backlog issues should remain open unless explicitly marked obsolete or superseded.

itachi-hue:
Closing this broad P2 audit grab-bag as stale/not actionable. Individual concrete items have since been split/fixed in focused issues and tests (for example #1562, #1717, #1703, #1784, #1517/#1877). Remaining polish should be filed as focused issues with current screenshots.

PR #1611, opened 2026-06-20
fix(workers): card status badge reflects real state + label success-rate (#1565)

federicodeponte:
Fixes #1565 P2-1 and P2-2 (worker-card polish).

## P2-1 — "Draft" on every card despite "5 active"
Root cause: the header count and the per-card badge used **two different axes** for what reads as the same concept.
- Per-card badge = **stage** axis (`workerStageKey`: Draft vs Live). A plain worker with no explicit `stage` defaults to `draft`, so most cards show "Draft".
- Header "active" count = **health** axis (`status === healthy|ready`).

So "5 active" (health) could sit next to a grid where every card shows "Draft" (stage) — they never had to agree.

Fix: count **live (non-draft)** workers on the *same axis as the badge* and relabel the header chip `active → live`. Now the header number equals the number of cards **without** a Draft badge. The "needs attention" health chip is untouched (it's a separate, non-contradictory signal). Added `workerStageLabel` helper + unit test.

## P2-2 — unlabeled success rate
The card meta line showed a bare `29%` after `318 runs`. Labeled it `29% success` so it's unambiguous, kept compact.

## Verification
- `npx tsc --noEmit` → 0
- `npx next build` → 0
- `npm run lint` → 0 errors
- `npx vitest run` → 576 pass; only the pre-existing `login-secrets` (`api.me is not a function`) failure remains, unrelated to this change.


itachi-hue:

Issue #506, opened 2026-06-07
FL10: Clear all runs is too prominent near search


federicodeponte:
## Description
Clear all runs sits by the search bar, making a destructive action too prominent and easy to trigger.

## Evidence
Source: /tmp/fede-livetest-issues-2026-06-07.md, FL10.
Fix PR: #493.

## App
OSS.

## Expected
Dangerous destructive actions are guarded and not one-click prominent near routine controls.

federicodeponte:
Wave 1 deploy verification note, 2026-06-07: this issue was auto-closed by PR merge keywords. The merged web bundle was built successfully, deployed to https://workers.floom.dev, and included in the post-deploy route smoke with no 5xx/508 responses. Additional explicit live evidence captured during this wave includes /assistant rendering with the updated sidebar order and editor chrome in /root/ax-browser-broker/artifacts/screenshots/pool-b-1780807401426.png.

PR #1619, opened 2026-06-20
perf: resolve only declared run secrets

itachi-hue:
## RCA

The remaining perf hypothesis was duplicate second secret resolution before E2B env injection. Code review did not show a literal second plaintext resolve on the normal execute_run path: execute_run resolves run_secrets once and passes that same dict into the E2B driver for .env.local, secrets.json, command envs, and log scrubbing.

But there was a real hot-path inefficiency in the same area: get_secrets_for_worker listed every DB secret name for the owner/workspace and resolved the union of declared worker secrets plus all available DB secrets. On secret-heavy workspaces, that means unnecessary name enumeration and unnecessary Vault decrypt work for secrets the worker did not declare or need.

That also weakens the documented sandbox contract: the function doc says the sandbox should receive only secrets declared by worker.yml.

## Fix

- Resolve only config.secrets for worker run injection.
- Keep the existing platform-secret denylist.
- Filter repository results back to the declared allowlist before caching/returning, so an over-broad repository implementation cannot accidentally ship undeclared secrets.
- Update hot-path cache test to assert list_names is not called and undeclared secrets are not returned.

## Expected latency impact

For NovaSearch, this removes the extra workspace-wide list_names query and prevents decrypting unrelated workspace secrets. After cloud PR #506, declared secrets are already batched; this keeps the batch as small as the worker manifest requires.

This will not remove E2B API/sandbox latency, but it should reduce remaining pre-sandbox work in secret-heavy workspaces.

## Tests

- python -m pytest apps/api/tests/test_executor_split_hotpath.py apps/api/tests/db/test_scheduled_run_secret_scope.py -q -> 10 passed
- python -m pytest apps/api/tests/test_openai_platform_secret_split.py -q -> 12 passed
- python -m ruff check apps/api/services/run_secrets.py apps/api/tests/test_executor_split_hotpath.py -> passed

Additional note: apps/api/tests/test_workspace_donation_and_token.py currently fails on unrelated admin/workspace-token setup errors, 403/404 before the secret assertion. I did not change those paths.


itachi-hue:
Reviewed — approve. Removes the workspace-wide `_secret_names_from_db()` enumeration, resolves only the declared allowlist and filters the result back to it; the `UNDECLARED_API_KEY` test proves undeclared secrets don't ship. Bigger value is least-privilege (runs only get declared secrets); latency gain is modest post-#506 (already 1 batch RPC), as noted. Watch-item on deploy: any worker relying on an *undeclared* workspace secret will now fail 'missing credential' — fix is to declare it, not revert. Merging + deploying + testing NovaSearch now.

Issue #1620, opened 2026-06-20
[perf] Opt-in: bake worker bundle into a per-worker E2B template to skip cold-start bundle upload


itachi-hue:
## Summary

Add an **opt-in, per-worker option to bake the worker bundle into a dedicated E2B template** so that **cold-start runs skip the per-run bundle upload**. The toggle must be **declarative config in `worker.yml`** (e.g. `exec.bundle_baked: true`), default **off** — no worker code change, mirroring the existing `WORKEROS_E2B_PYTHON_DEPS_BAKED` deps-baking precedent.

## Motivation / measured evidence

On the cloud NovaSearch worker (E2B, 65-file bundle = `run.py` + `lib/` + `data/`), the per-run **worker-bundle + context upload to the E2B sandbox is ~7–10s**. After the warm-pool (#1618) and batch-Vault (cloud #506) fixes landed, this upload is now one of the largest remaining chunks of the **cold-start** boot time.

Crucially, this cost is **cold-start only**. The driver already skips the bundle upload on warm-pool reuse:

```python
# engine/apps/api/runner_sandbox/e2b_driver.py:1801
if not sandbox_prepared:                 # sandbox_prepared == leased from warm pool
    upload_tree_tarball(... label="worker bundle" ...)   # bundle upload only on cold spawn
```

So a warm run pays nothing for the bundle; a **cold** run (first run after a deploy, or after the pool ages out at `WORKEROS_E2B_WARM_POOL_MAX_AGE_SECONDS`, default 900s) re-uploads the full bundle every time. For **sporadic-use workers** (e.g. a recruiter firing one search every few hours), effectively **every run is cold**, so this upload is paid on every run.

Baking the bundle into the template removes that upload from the cold path entirely: the sandbox already contains the worker code at spawn.

## Why this is its own option (not default, not deps-baking)

- **Bundles change far more often than pip deps.** `WORKEROS_E2B_PYTHON_DEPS_BAKED` is safe to leave on because requirements rarely change. The bundle changes on **every worker code edit**, and baking requires an **E2B template rebuild (~1–3 min)** on each change. Default-on would punish every `workers_write_file` / deploy of a worker under active development.
- Therefore: **opt-in per worker, default off.** Stable, high-volume prod workers (NovaSearch) flip it on once and win on cold starts; dev/iterating workers stay on the fast upload path.

## Proposed design

[...]

PR #1621, opened 2026-06-20
feat: support opt-in E2B bundle-baked templates

itachi-hue:
## RCA

#1620 is about the remaining cold-start cost after warm-pool and batch-secret fixes: cold E2B runs still upload the read-only worker bundle into each fresh sandbox. Warm-pool reuse already skips this, but sporadic workers whose pool has aged out still pay the upload cost.

The runtime already had content-addressed template-cache plumbing, but it was not tied to a declarative worker opt-in and the driver still uploaded the bundle even when a worker-specific template was selected.

## Fix

- Add `exec.bundle_baked: true` to the 0.3 manifest model and project it into `WorkerRuntime.bundle_baked`.
- Make worker-template cache usage opt-in and globally disableable with `WORKEROS_E2B_BUNDLE_BAKED_ENABLED=0`.
- Return `(template_id, bundle_baked)` from E2B template selection so the driver knows whether the template actually contains the bundle.
- Skip cold-run worker bundle upload only when a matching baked template is selected.
- Keep fallback safe: if the worker opts in but no matching cache entry exists, the run logs a warning and uses the normal upload path.
- Add `apps/api/runner_sandbox/e2b_bundle_template.py` to stage bundles with the same run-scoped/cache exclusions and update the template-cache file.
- Add `ops/e2b/build-worker-bundle-template.py` for deploy automation.
- Update authoring and E2B ops docs.

## Rollout

Worker manifest:

```yaml
exec:
  bundle_baked: true
```

Build/update the cache:

```bash
E2B_API_KEY=e2b_... python ops/e2b/build-worker-bundle-template.py \
  --worker-dir workers/novasearch-v5 \
  --cache-file data/e2b-template-cache.json
```

Run the API with:

```bash
WORKEROS_E2B_TEMPLATE_CACHE_FILE=data/e2b-template-cache.json
```

Disable globally if needed:

```bash
WORKEROS_E2B_BUNDLE_BAKED_ENABLED=0
```

## Tests

- `python -m pytest apps/api/tests/test_e2b_template_config_1427.py apps/api/tests/test_e2b_declared_secret_env_concurrency.py -q` -> 27 passed
- `python -m pytest apps/api/tests/test_1433_context_mount_selection.py apps/api/tests/test_manifest_resilience.py -q` -> 21 passed
- `python -m pytest apps/api/tests/test_e2b_template_config_1427.py apps/api/tests/test_e2b_declared_secret_env_concurrency.py apps/api/tests/test_1433_context_mount_selection.py apps/api/tests/test_manifest_resilience.py -q` -> 48 passed
- `python -m ruff check apps/api/models.py apps/api/runner_sandbox/e2b_driver.py apps/api/runner_sandbox/e2b_bundle_template.py apps/api/tests/test_e2b_template_config_1427.py apps/api/tests/test_e2b_declared_secret_env_concurrency.py ops/e2b/build-worker-bundle-template.py` -> passed

## Notes

This PR does not add an always-on background template builder inside the API process. The safe production path is deploy automation: update the worker bundle, run the builder, then deploy/run with the cache file. Runs remain successful while the baked template is missing or stale because the driver falls back to normal bundle upload.


itachi-hue:
Follow-up from enabling this on cloud NovaSearch (2026-06-20): the **builder `ops/e2b/build-worker-bundle-template.py` doesn't run with the current `e2b` SDK** — `build_worker_bundle_template` passes absolute paths from `staged.iterdir()` to `TemplateBuilder.copy()`, which raises `TemplateException: absolute paths are not allowed`. Fix: build with `Template(file_context_path=str(staged))` + relative item names. Also the produced cache_key only matches at runtime if the build runs in the **runtime's Linux env** (the `_hash_tree` byte hash differs across OS/materialization). Full field notes + the ~2.6s cold-only payoff measurement are on #1620. The opt-in + safe-fallback design itself worked well.

Issue #1608, opened 2026-06-20
[security] Member PAT creation is not audit-logged despite #949 PR claim (follow-up to #949)


itachi-hue:
Closed issue #949 (https://github.com/floomhq/workeros/issues/949, PR #962) addressed 'members can create non-expiring API tokens'.

**What landed (credit):** every PAT now has a bounded TTL — `create_token` applies `_enforce_token_ttl_cap` (apps/api/routers/auth.py:419 -> services/auth_ops.py:396-415, 90d default, 422 beyond WORKEROS_PAT_MAX_TTL_DAYS), and a disabled/deleted member's tokens are revoked/rejected (#915).

**Deliberately not done (by design):** the issue's 'restrict creation to admin OR owner approval' recommendation was consciously rejected — the maintainer states PATs are the supported member CLI/MCP auth and gating them on admin would break that flow. The account-outlives-token risk is instead closed via TTL + lifecycle. This part is not a gap.

**Remaining gap:** the PR comment claimed 'creation is audit-logged with actor and expiry', but the member-PAT create path has NO audit/log line. `create_token` (apps/api/routers/auth.py:408-448) does not emit any logger/audit record. (Note: the *workspace* token path does log — 'workspace token created' at apps/api/routers/workspace.py:1154 — but that is the admin-only `wst_` workspace token, not the `wos_` member PAT #949 covers.)

**Acceptance criteria:** emit an audit/log record on `POST /auth/tokens` creation including actor user_id, token name, and expires_at, consistent with the existing workspace.py:1154 / secrets-mutation logging pattern.

itachi-hue:
**Scope correction (automated triage).** The original filed scope overreaches: part of it reflects a deliberate later decision or out-of-repo work, leaving a smaller genuine remainder.

Filed issue's headline ask (a: restrict token creation to admin/owner-approval) is a REMNANT — explicitly rejected in the closing comment: members keep PAT creation because PATs are the supported member CLI/MCP auth; the real risk is closed via bounded TTL (done) + #915 verification-rejection of disabled-member tokens. True remainder is only (c) the missing audit-log line, which is low-value defense-in-depth on already-bounded, lifecycle-protected tokens.

[...]

Issue #1607, opened 2026-06-20
[frontend-parity] Workspace settings: fallback_model missing and workspace max_output_tokens not applied as run default (follow-up to #797)


itachi-hue:
Closed issue #797 (https://github.com/floomhq/workeros/issues/797) asked for workspace-level model defaults + run limits: default_model, fallback_model, max_tokens_per_run, default_timeout_seconds, monthly_spend_cap_usd, and a current-spend readout.

**What landed (credit, PR #991):** default_model resolution in the agent driver (apps/api/runner_sandbox/agent_driver.py:66, consumed at :494); default_timeout_seconds ceiling (apps/api/run_service.py:1195, agent_driver.py:99); monthly_spend_cap_usd enforced via services/run_cost.py:63; current_month_spend_usd returned from GET /workspace/settings.

**Remaining gaps:**
1. **fallback_model is entirely absent.** `grep -rn fallback_model apps/api` (excluding tests) returns zero hits; it is not in `_WORKSPACE_SETTING_ALLOWED_KEYS` (apps/api/routers/workspace.py:1340-1346) and there is no runner-side fallback resolution. The Settings > System 'Fallback' field has no backend. (A prior maintainer comment folded fallback into the single default-model field; if a distinct fallback is still wanted per the wireframe, it needs a real key + resolution path.)
2. **Workspace-level max_output_tokens is stored but never consumed.** It is an allowed/validated key (apps/api/routers/workspace.py:1346, validated at :1404) but no code reads it as a per-run default: `_ws_default_int` is called only with 'default_timeout_seconds' (agent_driver.py:99), and effective max-tokens derives solely from per-worker `int(limits.max_output_tokens)` (agent_driver.py:446, :499). So a workspace default for max tokens/run has no effect.

**Acceptance criteria:**
- If a distinct fallback model is desired: add `fallback_model` to allowed keys + validation, and consume it in agent-driver model resolution (worker > workspace default > workspace fallback > system default).
- Apply workspace `max_output_tokens` as a per-run default when the worker declares none (precedence: per-worker limits.max_output_tokens > workspace max_output_tokens > system cap), with a test asserting the workspace value is honored.

itachi-hue:
[...]

Issue #1598, opened 2026-06-20
[security] GET /secrets has no server-side role gate + no read audit trail (backend remainder of #943)


itachi-hue:
Closed issue #943 (https://github.com/floomhq/workeros/issues/943) acceptance: (1) /connections/secrets restricted to owner/admin roles, (2) non-owner views hide/group secret names + worker mappings, (3) access to secret metadata is auditable.

What landed (credit, PR #965): AC#2 + client-side AC#1. The page is owner/admin-gated client-side via apps/web/lib/use-is-admin.ts and members make zero list calls (apps/web/app/connections/secrets/page.tsx; test apps/web/tests/secrets-role-gate-943.dom.test.tsx).

Remaining gap (my evidence on origin/main @0e83d63f), in-repo (apps/api):
- AC#1 server side: GET /secrets (apps/api/routers/secrets.py:219-270 list_secrets) has NO admin/owner role check — it answers any authenticated member scoped only by user_id, returning secret names + per-worker used_by mappings (:258-267). secrets_router (secrets.py:59) declares no router-level dependencies; only the mutation helper checks auth.is_admin (secrets.py:49). A member can call the API directly and bypass the client-side UI gate.
- AC#3 audit: the read path logs nothing; logger.info is emitted only on upsert (secrets.py:155) and delete (secrets.py:188).

The g2 closing comment explicitly deferred this to the API batch (g1/g3), but #943 was closed COMPLETED.

Acceptance criteria:
- GET /secrets enforces owner/admin role server-side (returns 403 for members), with single-tenant/no-role deployments preserving access.
- Reads of secret metadata are audit-logged (user, role, timestamp).
- Test coverage for member 403 + audit emission on the read path.

itachi-hue:
PR #1624: https://github.com/floomhq/workeros/pull/1624\n\nRCA: the UI hid Secrets from members, but GET /secrets itself had no server-side role gate and the read path emitted no audit event.\n\nFix: GET /secrets now returns 403 for non-admin roles while preserving role-less single-tenant access, and logs metadata reads with actor + role.\n\nTests: member 403 + audit-log unit coverage added. Focused Python batch passed: 73 passed; ruff passed.

itachi-hue:
Fixed by merged PR #1624 (ea8b8d7). RCA/fix/test details are in the PR and prior issue comment; closing now that it is on main.

Issue #1592, opened 2026-06-20
[agent-interface][P3] Rate-limit 429 responses lack structured error_code envelope (follow-up to #601)


itachi-hue:
Follow-up to #601 (https://github.com/floomhq/workerOS/issues/601, closed via #606).

## What landed (credit)
Per-IP auth limits and the run-create/chat quota are correctly in place on origin/main. The exact repro from the issue (40 rapid GET /auth/me with no 429) is now covered: /auth/me is rate-limited (30/60s), and login/setup/magic-link/tokens are in RATE_LIMIT_RULES.

## Remaining gap on origin/main (0e83d63f)
The 'clear 429 envelope' the issue asked for is the default FastAPI `{"detail": "..."}` shape with a `Retry-After` header and NO machine-readable code. Agents must string-match the human-readable `detail` text to detect rate-limit 429s.
- Middleware: apps/api/main.py:1517-1519 -> `JSONResponse(429, {"detail": "Rate limit exceeded"}, headers={"Retry-After": ...})`
- Run-create quota: apps/api/services/quota.py:95-97 -> `HTTPException(429, detail="Run creation rate limit exceeded: 10/60s", headers={"Retry-After": ...})` (chat at :198-200)
- Login lockout: apps/api/routers/auth.py:150-151 -> `HTTPException(429, detail="too many failed login attempts; try again later")`
- No custom 429 handler exists: `grep -rn 'rate_limit_exceeded|RateLimitExceeded|_rate_limit_exceeded_handler' apps/api` returns nothing.

The 402 spend-cap path already models the desired shape: apps/api/main.py:4151 `detail={"error_code": "spend_cap_exceeded", "message": ...}`.

## Acceptance criteria
- All 429 paths (middleware, quota.py run/chat, auth login-lockout) return a structured body with a stable `error_code` (e.g. `rate_limit_exceeded`) plus `message`/`retry_after`, keeping the `Retry-After` header.
- A test asserts the `error_code` field on a rate-limited response.
Effort: S.

itachi-hue:
[...]

PR #1624, opened 2026-06-20
fix: close backend leftover issues

itachi-hue:
## Summary

Fixes the non-UI/backend leftover batch except #1201, which was intentionally left alone.

## Issues covered

- #1588: fail OSS single-tenant startup when `FLOOM_SECRET` is missing outside explicit `WORKEROS_DEV=1`; cloud/multi-tenant path remains exempt.
- #1598: enforce server-side admin/owner access for `GET /secrets` and audit-log metadata reads.
- #1608: audit-log member PAT creation with actor, token name, and expiry.
- #1592: return structured `rate_limit_exceeded` envelopes for middleware, run/chat quota, and auth lockout 429s while preserving `Retry-After`.
- #1607: add validated `fallback_model` workspace setting and apply workspace `max_output_tokens` as the default when the worker uses the sentinel/unset limit.
- #1581: fix invalid `test_local-user...` test name so `test_local_workspaces.py` collects.
- #1579: add deterministic worker-author prompt-derived `suggested_id` fallback and a five-prompt distinct-id regression test.
- #1526: make `workers validate` catch server-parity schema failures for `use_cases` count and non-string placeholders.
- #1575: make `workers push` explain hidden global/cross-workspace worker-id conflicts when GET 404 is followed by POST 409.

## Validation

- `python -m pytest apps/api/tests/test_998_signer_fallback_removed.py apps/api/tests/test_local_workspaces.py apps/api/tests/test_perf_fast_paths.py apps/api/tests/test_1592_1608_rate_limit_and_pat_audit.py apps/api/tests/test_workspace_settings_794.py apps/api/tests/test_797_workspace_defaults_enforcement.py tests/test_pr_186_draft_id_collision.py -q` -> 73 passed
- `npm test -- --runTestsByPath test/cli-workers.test.js` in `apps/mcp` -> 74 passed
- `python -m ruff check ...` on touched Python files -> passed


itachi-hue:
## Review — clean, well-tested backend batch. Approve, with one verification on #1598.

Scope is all app/test/worker code; tests included; no hardcoded secrets.

**Verified ✓**
- **#1588** FLOOM_SECRET fail-closed startup — correctly gated to OSS prod only (`deploy=="local"` and not `WORKEROS_DEV=1`); **cloud exempt** (cloud strips `FLOOM_SECRET`), dev exempt. Good hardening.
- **#1592** structured 429s — adds `error_code: "rate_limit_exceeded"` + `retry_after` while **preserving the `Retry-After` header** (tested: `assert response.headers["Retry-After"]`).
- **#1608** PAT audit — logs actor/role/token-name/expiry, **not** the token value. Correct.
- **#1607** `fallback_model` — routed through the same safe model-id validator as `default_model`. No injection.
- #1581 / #1579 / #1526 / #1575 — low-risk dev-ergonomics/validation fixes, with tests.

**Flags**
1. **⚠️ #1598 — confirm a workspace OWNER passes `is_admin`.** The new gate is `if auth.role and not auth.is_admin: 403`, but `is_admin` only covers `role == "admin"` / `"admin"` scope — **not `role == "owner"`**:
   ```python
   @property
   def is_admin(self) -> bool:
       return self.role == "admin" or "admin" in self.scopes
   ```
   If cloud owners carry `role == "owner"` without an admin scope, they'd get `403 "admin role required"` on their **own** `GET /secrets`. OSS single-tenant is unaffected (role `None` → gate skipped). The existing `_require_admin` has the same shape, so owners likely already get `is_admin` via a mapping/scope — but since this PR is specifically about secret access, please confirm with a test using an **owner** actor (not just admin).
2. **#1588 is a behavior change** for self-hosters running OSS prod (`deploy=local`) without `FLOOM_SECRET` and without `WORKEROS_DEV=1` — startup now fails (intended hardening). Worth a changelog/release note so misconfigured deploys aren't surprised.
3. CI red = inherited `main`-red (#1201) + Vercel author-permissions, not this PR.

Only thing I'd actually gate on is #1598 not locking owners out of their own secrets.

_Automated review (Claude Code)._

itachi-hue:
Addressed the #1598 review flag in the PR branch.

RCA: the backend `AuthContext.is_admin` predicate treated only `role == admin` or `admin` scope as admin. That meant a cloud/workspace owner represented as `role=owner` could be blocked by the new `/secrets` role gate even though frontend and workspace semantics already treat owners as admins.

Fix:
- `AuthContext.is_admin` now treats `role in {owner, admin}` as admin.
- Added a regression test that `GET /secrets` allows an owner actor and still rejects members.

Validation after the update:
- `python -m pytest apps\\api\\tests\\test_perf_fast_paths.py -q` -> 13 passed
- `python -m pytest apps\\api\\tests\\test_perf_fast_paths.py::test_list_secrets_rejects_non_admin_roles apps\\api\\tests\\test_perf_fast_paths.py::test_list_secrets_allows_workspace_owner apps\\api\\tests\\test_perf_fast_paths.py::test_list_secrets_audit_logs_metadata_read -q` -> 3 passed
- `$env:WORKEROS_DEV='1'; python -m pytest apps\\api\\tests\\test_multi_member.py -q` -> 32 passed, 1 existing Starlette warning

Note: running `test_multi_member.py` without `WORKEROS_DEV=1` now fails at startup due to the intentional #1588 `FLOOM_SECRET` hardening; with explicit dev mode, the suite is green.

PR #1675, opened 2026-06-20
fix(analytics): trustworthy post-hoc metrics (duration outliers, failure taxonomy, token capture)

federicodeponte:
## Problem (verified live on workeros-api.floom.dev, runner e2b)

1. **Duration poisoned by orphaned runs.** `/stats` returned `avg_duration_ms: 30,819,058` (8.5h); `/workers/morning-brief/stats` returned `avg: 60,012,999`, `p95: 52,431,881`. Culprit: `run_ee90e0db5937` (failed, `duration_ms: 52431881` = 14.5h). A single stuck run makes mean + p95 meaningless.
2. **`total_tokens` / `total_cost_usd` null on every run.**
3. **No failure taxonomy.** ~25% 7d success rate; failures surface only as the last raw `error` string.

## Root cause

1. **Duration:** `duration_ms` is correctly `completed_at - started_at` (sqlite + cloud Supabase repo). 52431881ms is a *real* wall-clock orphan (hung sandbox / dead drain loop / late reaper). The bug is the **aggregation** in `worker_telemetry.py` — plain mean + naive p95, no outlier handling. The absolute timeout ceiling is `MAX_RUN_TIMEOUT_SECONDS` (1h), so anything far above it is provably an orphan.
2. **Tokens:** `services/run_cost.py::_persist_run_cost` wrote via raw `get_db()` (**sqlite only**). On the cloud the data is in Supabase, so the write hit an unused local file. `RunDetail.total_tokens` is separately computed from the transcript at read time (`routers/runs.py`).
3. **Taxonomy:** `error_code` often null; no classifier, no breakdown field.

## Fix (minimal, public engine)

- **New `services/run_metrics.py`** (pure, tested):
  - `duration_summary()` drops durations above `MAX_RUN_TIMEOUT_SECONDS * DURATION_OUTLIER_TIMEOUT_FACTOR`, adds a robust **median**, reports `duration_outliers_excluded`.
  - `classify_failure()` / `failures_by_category()` map `error_code` (preferred) or the raw message to: `timeout / crash / config / auth / network / resource / validation / quality / cancelled / unknown`. Message fallback handles the cloud where `error_code` is mostly null.
- `routers/worker_telemetry.py`: both stats endpoints use the helper; new `median_duration_ms`, `duration_outliers_excluded`, `failures_by_category`.
- `services/run_cost.py` + `run_service.py`: `_persist_run_cost` routes the token/cost write through `repos.runs.update(...)` (works on sqlite AND cloud Supabase); sqlite fallback kept for legacy callers.
- `db/sqlite.py`: add the two columns to the runs `update` allow-list.
- `models.py`: new optional fields on `WorkerStats` / `WorkspaceStats`.

**Trade-off:** threshold-clamp at the absolute timeout ceiling (not IQR/MAD) — principled (a run cannot legitimately exceed the ceiling), no minimum sample size, never drops a real slow run. Mean + p95 kept (now clean); median added for a robust headline.

## Tests

- `tests/test_run_metrics.py` (20 cases): stuck-run-must-not-poison fixture (uses the live 52431881ms value), boundary, all-outliers, taxonomy by code + message + precedence.
- `tests/test_monitoring_apis.py`: `test_stuck_run_does_not_poison_avg_duration`, `test_failures_by_category_is_present`.
- `tests/test_surface_cost_nextrun_r9.py`: `test_repo_update_persists_tokens_and_cost`.
- Full `test_monitoring_apis.py` + `test_run_metrics.py`: **71 passed**; the only 2 failures (`TestEmailNotifications`, `TestOverviewConsecutiveFailureAlerting`) are **pre-existing on clean `origin/main`** (verified by stashing the diff), unrelated.

## Scoped as follow-up

- **Cloud column + allow-list:** done in `floomhq/workeros-cloud` `fix/posthoc-analytics-token-capture` (migration 0048 + supabase_repos allow-list). **Land this engine PR + bump the submodule first**, then the cloud change.
- **Cloud spend aggregate:** `_worker_month_to_date_cost_usd` still reads raw sqlite; same bug class, left out to keep this PR minimal. The persisted column now exists for it.
- **Cloud transcript availability:** confirm cloud e2b runs materialize `transcript.jsonl` to disk so read-time `total_tokens` extraction works (the driver emits the `usage` row only when `total_tokens > 0`).


itachi-hue:
Fully superseded by #1756 (and the parallel analytics work). Verified mechanically: rebasing this branch onto current `main` produces a **zero-line diff** (identical tree hash `d7a5af56`). Everything here — the failure-taxonomy refinements + duration-outlier metrics in `run_metrics.py`, the `run_cost.py` repo-path write + trace-derived cost, and the `worker_telemetry.py`/`models.py`/`sqlite.py` changes — is already in `main`. Nothing left to merge, so closing. The work landed, just via a different PR.

PR #1686, opened 2026-06-20
fix(ui): worker Source view single-click + correct active-view checkmark (#1680)

### LINKED ISSUE #1680: [P2] 'View as YAML'/Advanced→Source needs a 2nd click to render; Advanced menu shows two checkmarks at once
Advanced→Source and 'View as YAML →' both require a second click before the Source view actually renders. The Advanced dropdown also shows two checkmarks simultaneously (Source + Versions), so the active-view state is wrong. evidence /tmp/screen-scores/.

federicodeponte:
Closes #1680.

## Symptoms
- `View as YAML →` (Setup tab) and **Advanced → Source/Versions/Brain/Tools** required a **second click** before the view rendered; the first click did nothing visible.
- The Advanced dropdown showed **two checkmarks at once** (Source + Versions) — it was lighting up the *pinned set*, not the *active view*.

## Root cause
**Double-click:** `Collection` (the URL-bound wrapper) holds view state in local React state, seeded from the URL on mount and re-seeded only on `popstate`. The `View as YAML` link and the Advanced view switcher navigate via the **Next router** (`router.replace`), which updates `useSearchParams` but never fires `popstate` and never touches that local state — so the view didn't switch until something else re-rendered/re-seeded it. In-component tab clicks worked because they go through `patch()` (local state + `window.history.replaceState`); only the router-driven entry points were broken.

**Double checkmark:** `CustomizeTabsMenu` rendered `checked={pinned.has(key)}`. Pinning is a persistent set, so every pinned advanced tab showed a checkmark. The menu reads as a view switcher, so its checkmark must track the **active view** (mutually exclusive).

## Fix (minimal, root-cause)
- **`Collection.tsx`**: re-seed local state from `useSearchParams` whenever it changes. In-component edits commit via `window.history.replaceState` (which does **not** update `useSearchParams`), so the effect reacts only to genuine router navigations — no loop, no clobber of local edits. Single click now switches the view in one render. Deep-linking (`?sel=…&tab=…`) is preserved (already URL-seeded; the new effect strengthens it).
- **`types.ts` / `CollectionView.tsx`**: thread the resolved active-tab key into `config.detail(item, activeTab?)` (optional 2nd arg; the other 6 collections ignore it — no break). `config.detail` is pure, so the two-pass call is side-effect-free.
- **`WorkersCollection.tsx`**: the Advanced menu checkmark is now `key === activeTab` → exactly one (or none) is ever checked. Selecting an advanced tab opens it (pinning it onto the row first if needed); re-picking the active tab unpins it, preserving the only unpin path.

## Verification
- `npx tsc --noEmit`: 0 new errors (only the pre-existing `command-palette.test.ts` baseline error).
- `npm run lint`: 0 errors.
- `npx next build`: ✓ compiled successfully.
- `npx vitest run`: **596/596** pass, including **2 new** `#1680` tests asserting the menu marks only the active advanced view (mutually exclusive, never two checkmarks).


itachi-hue:

PR #1691, opened 2026-06-20
fix(ui): per-worker Runs tab uses correct run filter (#1679)

### LINKED ISSUE #1679: [P1] Per-worker Runs tab shows 'No runs yet' while global Runs shows 200 for same worker (data inconsistency)
Embedded per-worker Runs tab renders 'No runs yet' while the global /runs page shows 200 runs for the SAME worker; the 'Runs N' badge flips 0↔1 across tab clicks. The embedded tab is querying with a wrong/missing filter. evidence /tmp/screen-scores/03*.

federicodeponte:
Closes #1679.

## Root cause (frontend)
The embedded per-worker **Runs tab** and its **"Runs N" tab badge** both read the worker-detail `recent_runs` field. The backend builds that field via `runs.list_for_worker`, whose SQL is scoped **only** by the worker owner:

```sql
WHERE w.owner_id = ? AND r.worker_id = ?
```

The global `/runs` page instead calls `api.runs.list({ worker_id })` → `_list_visible_runs` → `runs.list` / `list_operator_visible`, which is scoped by **actor OR owner**:

```sql
WHERE (COALESCE(r.actor_user_id, w.owner_id) = ? OR w.owner_id = ?) AND r.worker_id = ?
```

So for a worker whose runs were triggered by a non-owner actor, the narrow per-detail query returns **empty** while `/runs` shows the full history (the reported 200) for the same worker — hence "No runs yet" in the tab.

The **badge flipped 0↔1** for the same reason: before the worker detail loads it fell back to `w.last_run ? 1 : undefined` (1), then settled to the empty `recent_runs.length` (0).

This is a **frontend** bug: the proven-correct broad query already exists and works — the embedded tab just wasn't using it.

## Fix
- `RunsTab` now sources its rows from `api.runs.list({ worker_id })` — the **same** query the global `/runs` surface uses — instead of `recent_runs`.
- A module-level `workerRunsCache` (mirroring the existing `detailCache` pattern) is the single source of truth for **both** the tab body and the synchronously-rendered badge count, so the two can no longer disagree and the badge settles monotonically (fallback → real count) rather than oscillating 0↔1.

## Verification
- `npx tsc --noEmit`: 0 new errors (one pre-existing `command-palette.test.ts` error unrelated to this change, present on `main`).
- `npm run lint`: 0 errors.
- `npx next build`: ✓.
- New test `apps/web/tests/worker-runs-tab-filter-1679.dom.test.tsx` proves the tab queries `api.runs.list({ worker_id })` and renders runs even when `detail.recent_runs` is empty. Targeted worker/runs/collection suites: 30 passed.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>


itachi-hue:

Issue #1699, opened 2026-06-20
[P1] Cold-start home CTAs reference data a new user doesn't have (no zero-state on redesigned home)


federicodeponte:
**Frame:** first-run / onboarding audit 2026-06-20 (non-technical operator, first 5 min). Live: https://app-preview.floom.dev/app

## Problem
The redesigned "Good evening" Emily home is what a brand-new user lands on, but every teaching element on it assumes a populated workspace:
- Headline metric: "116 done this week · 4 need attention" — for a new workspace this is "0 done", a wall-of-zeroes greeting.
- Prompt chips: **"What ran overnight?"**, **"Show me this week's runs"** — both reference runs that do not exist for a new user. Only "Create a Linear triage worker" is useful cold.

There is **no first-worker guidance** on the screen a new user actually lands on: no "what would you like to automate?" composer, no templates, no teach of what a worker is.

## Root cause / note for fix
The repo has a proper zero-state — `apps/web/components/overview/ActivationPanel.tsx` ("What would you like to automate?" composer + 4 example templates) — but it is gated in the OLD `OverviewDashboard.tsx` on `active_workers_count === 0 && completedThisWeek === 0`. The **deployed redesigned home does not appear to render ActivationPanel** for a zero-state. The new Emily home needs an equivalent empty-workspace branch: swap the data-assuming chips + metric for the activation composer + templates when the workspace has no workers/runs.

## Relationship to existing issues
- Supersedes the premise of #1363 (which said the Emily dock is *passive* on first load) — the redesigned home is now active but shows the WRONG content for a first-timer.
- Related to #1586 (first-session funnel) and #1092 (prompt box not just Hire), but those predate the home redesign; this is specifically "the new home has no zero-state branch."

## Evidence
- `docs/launch-readiness/agent-runs/onboarding-evidence-20260620/01-app-landing.png`

UI-lane

itachi-hue:
Closing as stale/addressed in current main. The redesigned home now renders a zero-worker branch in `EmilyHomeEmpty` with the first-run headline "Let's hire your first worker", gated by the real workers query via `resolveWorkersGate`; it no longer shows only data-assuming run chips for a brand-new workspace.

Verified with `cd apps/web && npm test -- --run tests/emily-home-onboarding-first-run.dom.test.tsx tests/emily-home-fullscreen.dom.test.tsx tests/create-flow-home-fullscreen.dom.test.tsx` as part of the launch smoke.

PR #1714, opened 2026-06-20
fix(ui): reliable New-worker first action + zero-state home gate (#1698 #1699)

### LINKED ISSUE #1698: [P1] 'New worker' / ?create=1 is a dead-or-broken first action (onboarding)
**Frame:** first-run / onboarding audit 2026-06-20 (non-technical operator, first 5 min). Live: https://app-preview.floom.dev/app

## Problem
The single most important first action — create your first worker — is ambiguous and route-dependent, and fails hard from some entry points.

1. **Dead click from home:** clicking **New worker** while on the home (`/app`) re-lands on the same "Good evening" home with no modal, no focus change, no field highlight. For a first-timer nothing visibly happens. (The intended behavior is "focus the Emily composer," but there is zero visual feedback that a click did anything.)
2. **Hard error from a non-home route:** opening the same deep-link `/app?create=1` from another route (reproduced from `/app/runs`) fires an error toast — **"Couldn't open that item; it may have been deleted."** — and the create flow never opens. The deep-link appears to be resolved as "open worker by id" and fails.

## Why it matters
This is the primary cold-start CTA. A new user who clicks it either sees nothing happen (home) or an error toast (elsewhere). Time-to-first-value is blocked or confusing at the very first step.

## Distinct from existing issues
- #1663 = Emily *chat* create path yields no worker (backend/data + cache). This issue is the **entry affordance / deep-link routing** itself, upstream of the chat.
- #1586 = first-session funnel (template gallery, setup chips). This is the broken button, not the funnel content.

## Evidence
- `docs/launch-readiness/agent-runs/onboarding-evidence-20260620/07-create-worker.png` (toast error on Runs route)
- `docs/launch-readiness/agent-runs/onboarding-evidence-20260620/08-create-flow.png` (New worker from home → same home, no change)

## Suggested fix
`?create=1` must deterministically open the create/compose affordance regardless of current route, with visible feedback (focus + caret in the composer, or a modal). Never resolve `create=1` through the open-worker-by-id path.

UI-lane

federicodeponte:
Closes #1698, #1699.

## #1698 — "New worker" / `?create=1` is a reliable first action

**Root cause:** the create empty-state composer was never focused on entering create mode, so clicking **New worker** (on home, where the create empty state is visually identical to the home empty state for a new user) produced no visible change — a dead click. Separately, a create deep-link (`?create=1`) landing on a **non-home** route was never consumed: `createMode` and the param-strip both required `isHomeRoute`, so on e.g. `/runs?create=1` the param just sat in the URL doing nothing while the route's Collection kept owning the surface.

**Fix (root-cause, minimal):**
- `PromptInput` gains an `autoFocus` prop; the home/create centered composer passes `autoFocus={createMode}` so entering create lands a caret in the composer → immediate visible feedback from **any** route.
- The `EmilyDock` create effect now forwards `?create=1` landing on a non-home route to `/?create=1` (preserving `&prime=`), so the create flow opens **consistently** regardless of entry route, and the create param never lingers on a Collection route.

Note: I could **not** reproduce the literal *"Couldn't open that item; it may have been deleted."* toast from `create=1` against this branch — `parseCollectionState` only reads `sel`, and `create`/`prime` are never mapped to a selected id (the original false-not-found toast was hardened in #1558). The redirect above is the structural guarantee that the create param can never reach a Collection's item resolver from any route.

## #1699 — zero-state on the redesigned home

**Investigation result: the gate is CORRECT in code.** The redesigned home renders `EmilyHomeEmpty`, which gates on `resolveWorkersGate` (success-keyed, never tripped on error/loading). A workspace with **0 real workers** (and therefore 0 runs — runs belong to workers) renders the *"Let's hire your first worker"* hero with create pills, and does **not** show the `{done} done this week` populated greeting. A populated workspace keeps the normal home. The live *"116 done this week"* a new user saw was a populated/stale state, not a code gate bug. Per the brief, I'm reporting this rather than inventing a fix — and adding deterministic coverage so it can't regress.

## Tests
New `tests/emily-home-onboarding-first-run.dom.test.tsx` (follows the existing `emily-home-*.dom.test.tsx` pattern):
- empty workspace (0 workers/0 runs) → first-worker hero, no populated greeting
- populated workspace → normal home (not ripped out)
- create mode focuses the composer (`document.activeElement === composer`)
- create mode never fires the false toast

## Verify
- `npx tsc --noEmit` → 0 new errors (1 pre-existing `command-palette.test.ts` error only)
- `npx next build` → ✓
- `npm run lint` → 0 errors
- `npx vitest run` → 124 files / 598 tests pass (incl. the 4 new)


itachi-hue:

PR #1719, opened 2026-06-20
fix(approvals): standalone review page scrolls long approvals (clipped past one screen)

federicodeponte:
## Bug
The standalone Approvals review page (`/approvals/review`) **doesn't scroll** — any approval taller than one screen (e.g. a multi-item shortlist) is clipped and items past the fold are unreachable. Confirmed live on a 5-item approval: only ~2 items render, rest unreachable.

## Root cause
`app/layout.tsx` sets `<body class="h-screen overflow-hidden">` (correct — the dashboard's own panels scroll internally). But the review page used `min-h-screen` with a plain `flex-1` content area and **no** `overflow-y-auto`, so inside the height-locked body its content was clipped.

## Fix (2 classes, root + global)
- outer: `min-h-screen` → `h-screen` (lock to viewport so the inner region can own the scroll)
- content: add `min-h-0 overflow-y-auto` to the `flex-1` region

Header stays pinned; content scrolls. Fixes scroll for **any** long approval, not just candidate lists — no use-case-specific code.

## Deploy note
⚠️ The cloud engine is currently pinned to a **diverged snapshot** (`24272c67`, `round09-base-engine`) that is NOT an ancestor of `main`. This fix lands on `main`; the cloud must be repointed off that snapshot (or this cherry-picked) for it to reach production. Flagging separately — that's the bigger deploy-hygiene item.


itachi-hue:

PR #1746, opened 2026-06-21
fix(approvals): embed preview media inline and surface review link in CLI

### LINKED ISSUE #1732: [MEDIUM] API/CLI don't expose the standalone approval-page URL — agents can't surface 'approve here' links
When a worker run lands in `pending_approval`, there's a standalone approval page (e.g. `https://workeros.floom.dev/app/approvals/<approval_id>` and `/app/runs/<run_id>` — both return 307, real routes). But nothing in the API/CLI surfaces that URL:
- `runs show` / `runs show --json` returns `approval_trail.id` (apr_...) but **no approval_url / page link**.
- `runs` CLI has `approve`/`reject` but does not print where a human would review it.

So an agent that creates an approval can't tell the user where to approve — it has to dig the `apr_` id out of raw JSON and guess the route. Please:
1. Return an `approval_url` (and `run_url`) field on the run + approval objects.
2. Print 'Review/approve at: <url>' in `run`/`runs show` CLI output when status is pending_approval.
3. If a shareable/public standalone approval link exists (cf. #1508), expose that too.

Env: @floomhq/workeros 4.2.4, cloud, workspace Nova Search. Example: run_425597164db8 → approval apr_25f11ed598df → page https://workeros.floom.dev/app/approvals/apr_25f11ed598df (307, exists, but never surfaced by the tooling).

### LINKED ISSUE #1739: [FEATURE] Approval/review pages should EMBED video (and images) inline, not just show a URL
For a content-publishing pipeline, the human approves by **watching the video**, not by reading a link. Today the HITL approval/review surface renders `decision_required.preview` as **text** (and output `media_type` only special-cases text/markdown + text/csv per apps/api/models.py). So a video post shows up as a plain URL in the preview — you can't review it on the page.

Ask:
- Let `decision_required` (and the review-pack body) carry a media field — e.g. `preview_media: [{type:'video', url, mime, poster}]` — and have the approval page (`/app/approvals` + the public `/review/<token>`) render an inline `<video>` player (and `` for images).
- Auto-detect when a preview/output value is a video URL (.mp4/.mov, video/* mime) and embed it.

Use case: SF-vlog pipeline — the approval card for each short should play the 9:16 video inline so the reviewer can watch + approve in one place, across IG/TikTok/X.

Env: @floomhq/workeros 4.2.4, cloud. Related: #1732 (approval link not surfaced), #1508 (public review renders payload links).

federicodeponte:
Two focused, independent fixes for the approval / review flow.

## Fix 1 — embed video/images in the approval/review pages (Closes #1739)

The approval/review "Proposed output" rendered a worker's media URL as bare text, so a reviewer had to copy it into a new tab to actually watch the video. (Federico reviews videos by watching them on the approval page.)

- New `apps/web/components/share/PreviewMedia.tsx` helper scans the preview text for the first **safe** `http(s)` URL ending in a video extension (`.mp4/.mov/.webm/.m4v`) or image extension (`.png/.jpg/.jpeg/.gif/.webp`) and renders an inline `<video controls playsInline>` / ``, **in addition** to the existing text.
- Reused by `ProposedOutput` in `ApprovalReviewBody`, so the **in-app** approval body and the **public** `/approvals/review` surface render identically (both already consume `ApprovalReviewBody`).
- Safe by construction: only `http(s)` URLs pass the existing `sanitizeHref` allowlist (no `javascript:`/`data:`/`file:`), no autoplay (no surprise sound), `` carries `referrerPolicy="no-referrer"`. (`referrerPolicy` is not a valid `<video>` attribute in React, so it is applied only to ``.)
- Trailing sentence punctuation and query strings on the URL are handled.

## Fix 2 — surface the approval/review link via the CLI (Closes #1732)

`chat_service` already returns approvals with a review link, but the CLI `runs show` returned `approval_trail` **without** it.

- Lifted the tokenised `/approvals/review?id=..&token=..` URL builder into a single shared helper `core.approval_signing.try_approval_review_url` (DRY: the chat tool, the run-detail serializer, and the CLI now all emit the **same** link).
- Refactored `services/chat_approvals._tool_approvals_list_pending` to use it (removing its private copy of the host + token shape).
- Added `link` to the `ApprovalEntry` model and populated it on `GET /runs/{id}` (`approval_trail.link`), only while the run is pending. DEGRADE, never raise: `None` when no signer secret is configured (hosted mode), so the response never 503s.
- `runs show` now prints `Review/approve at: <link>` when status is `pending_approval`, and includes it in `--json`.

## How tested

- **apps/web:** `npx vitest run` → **601 passed (124 files)**. New `tests/approval-preview-media-1739.dom.test.tsx` (7) covers video/image embed, text-only fallback, and the `findPreviewMedia` extension/safety matrix. `tsc --noEmit` clean for changed files; `eslint` clean (0 errors).
- **apps/mcp:** `npm run build` clean; new `test/runs-show-link.test.js` (3) covers pending link print, non-pending omission, and `--json` passthrough. 30 passed across run/redaction/workers suites. (Two pre-existing `integration.test.js` failures are unrelated to this change — they reproduce on `main`.)
- **apps/api:** `pytest tests/test_approval_review_url_1732.py tests/test_approval_list_signer_degrade.py` + approval/run-share suites → **37 passed**. `ruff check` clean. (One pre-existing `test_chat_phase1_backend.py` failure is unrelated and reproduces on `main`.)

Each commit is signed off (DCO).


federicodeponte (commented):
Adversarial review — one blocking issue (the headline feature is broken in prod), plus nits.

## BLOCKER — CSP `media-src` blocks the embedded `<video>`, so #1739's main case (videos) silently fails

`apps/web/middleware.ts:172` sets:

```
media-src 'self' blob:
```

There is no `https:` in `media-src`. The new `PreviewMedia` renders `<video src="https://cdn.../final.mp4">` (`apps/web/components/share/PreviewMedia.tsx:62-73`), and that route gets the CSP via the middleware matcher (`apps/web/middleware.ts:281`, `/approvals/review` is an HTML route, not a static asset). So the browser **refuses to load the external video**: the reviewer sees an empty `<video controls>` player that never plays. The `<a>Open video</a>` child is NOT a load-failure fallback — browsers only show it when `<video>` itself is unsupported, so on any modern browser there is no visible fallback either.

Images work only by luck: `img-src 'self' data: blob: https:` (line 171) **does** include `https:`, so `` loads. That's why the issue is invisible in review — the img path looks fine while the video path (the actual stated motivation: "Federico reviews videos by watching them ON the approval page") is dead.

The DOM tests can't catch this — jsdom/vitest doesn't enforce CSP. And `apps/web/tests/csp-cache-926-945.test.ts` was not updated to assert `media-src` permits the media host.

**Fix:** add the media host(s) to `media-src` in `buildCsp`, e.g. `media-src 'self' blob: https:` (matching the existing `img-src` allowance — these are non-executing media fetches), and add an assertion in `csp-cache-926-945.test.ts` that `media-src` includes `https:` so this can't regress. If a tighter allowlist is preferred, scope it to the known CDN host via an env seam like the existing `CSP_EXTRA_CONNECT_SRC`.

## NIT — `<video>` Referer leak is real but unmitigated; weigh it against the fix above

The PR comment notes `referrerPolicy` isn't a valid React `<video>` attribute, so only `` gets `no-referrer`. The `<video>` request will send a `Referer: https://.../approvals/review?id=..&token=..` header to the (worker-controlled) media host, leaking the tokenised review URL to a third party. The token authorizes that single approval, but it is still a capability URL being handed to an external host. Consider wrapping the `<video>` in a referrer-suppressing mechanism (e.g. a `<meta name="referrer">`-scoped approach is not per-element; a small same-origin proxy or stripping the token from the page URL after token exchange would close it). At minimum, document the accepted risk.

## NIT — DRY goal incomplete: two surfaces still hand-build the same link

The PR's stated purpose is a single source of truth for the `/approvals/review?id=..&token=..` link, but `apps/api/services/chat_tool_cards.py:294` and `apps/api/chat_service.py:1556`+`chat_tool_cards.py:270` still construct it by hand from `_APPROVALS_BASE_URL` + `try_approval_public_token`. `try_approval_review_url` should be reused there too, otherwise the exact drift the PR set out to eliminate still exists in `chat_tool_cards`.

## What's solid
- `try_approval_review_url` / `_approvals_base_url` faithfully match the prior `_APPROVALS_BASE_URL` precedence + `https://localhost:3000` default; degrade-to-None (no 503) is correct and tested.
- URL extraction regex is robust: handles query strings (`.mp4?token=`), fragments, trailing punctuation, multiple URLs (first media wins), case-insensitive ext, and rejects `javascript:`/`data:`/`file:` and non-media via `sanitizeHref`. Verified against ~15 edge cases.
- Text rendering unchanged (JSX text + `GenericOutput`, no `dangerouslySetInnerHTML`); `PreviewMedia` is additive.
- CLI `link` populates pending-only, omits when not pending, passes through `--json`; tests assert real behavior.

federicodeponte:
Addressed the review feedback in f87210c7:

**1. BLOCKER — CSP blocked the embedded video.** Fixed: \`media-src\` in \`apps/web/middleware.ts\` \`buildCsp\` now includes \`https:\` (\`media-src 'self' blob: https:\`), mirroring \`img-src\`. So the new \`<video src="https://...mp4">\` from \`PreviewMedia.tsx\` loads instead of being refused. Added a regression test in \`apps/web/tests/csp-cache-926-945.test.ts\` that asserts \`media-src\` contains \`https:\` (plus \`'self'\` and \`blob:\`). Web vitest: 19/19 pass, including the new test.

**2. NIT — Referer leak on \`<video>\`.** Since \`referrerPolicy\` is not valid on \`<video>\`, the tokenised review URL would leak via Referer to the worker-controlled media host. Mitigated on the public signed-link page (\`apps/web/app/approvals/review/page.tsx\`) by injecting a document-wide \`<meta name="referrer" content="no-referrer">\` (alongside the existing noindex meta, same effect pattern). This covers the \`<video>\` request and every other subresource. Documented the residual + rationale in a code comment.

**3. NIT — incomplete DRY.** \`apps/api/services/chat_tool_cards.py\` no longer hand-builds the \`/approvals/review?...\` URL; it now uses the shared \`core.approval_signing.try_approval_review_url\` helper (single source of truth for host + token shape, with the same degrade-to-None behavior). Dropped the now-unused \`_APPROVALS_BASE_URL\` / \`try_approval_public_token\` imports from that call site.
   - Note on the \`chat_service.py:~1556\` reference: that exact line builds \`{base}/approvals\` (the approvals *listing* URL), not a tokenised \`/approvals/review\` deep link, so it was intentionally left unchanged. The chat-approvals review-URL path was already migrated to the shared helper in this PR (\`chat_approvals.py\`); \`chat_tool_cards.py\` was the remaining hand-built review URL.

**Tests:** apps/api \`test_approval_review_url_1732.py\` 4/4 pass; apps/web vitest 19/19 pass (CSP + PreviewMedia DOM). Remaining failures in the suite (pytest-asyncio missing, case-sensitivity assertions in unrelated workspace-agent/command-palette tests) are pre-existing and not touched by this change.

itachi-hue:
Folded into and superseded by the integration PR #1749 (engine UI batch — review + merge as one). Closing to consolidate; the branch is preserved, so this can be reopened if #1749 drops it. Review/merge happens on #1749.

Issue #1717, opened 2026-06-20
Approval review page doesn't scroll — multi-item 'Proposed output' is clipped, later items unreachable


federicodeponte:
On the deployed Approvals review page (`/app/approvals/review?id=...`), when an approval's proposed output is a long list (e.g. a 5-candidate shortlist as `items[]`), the page does **not scroll** and the 'Proposed output' column is clipped — only the first ~1.5 items render; the rest are unreachable. A reviewer cannot see or act on items past the fold.

**Repro:** open an approval whose decision input has 5+ list items → only the first 1–2 render, no scroll, large empty space below. Confirmed live on `apr_8c21e86a64e1` (5 candidates, only ~2 visible).

**Note:** the deployed layout is the two-column 'Your response | Proposed output' view, which differs from the 480px single-card layout at repo HEAD — confirm which is live. Fix: the page / proposed-output column must scroll (or grow) to show all items.

itachi-hue:
Verified fixed in current web code: standalone approval review pages render in an owned vertical scroll container so long proposed output is reachable. Regression passed: `tests/app-shell-standalone-scroll.dom.test.tsx`.

Issue #1716, opened 2026-06-20
Cloud: public approval/review share links 503 — FLOOM_SECRET unset on the deployed API (no no-login client link works)


federicodeponte:
On the deployed cloud API, the public approval review path `/approvals/public/{id}` (and the signed `/approvals/review?id=&token=` link it backs) returns **503** because `FLOOM_SECRET` is not configured on the Railway environment. `GET /approvals` also returns `public_link: null`.

**Impact:** the standalone Approvals review page can only be opened by a **logged-in owner**. There is currently **no working no-login signed link** for an external reviewer on the cloud — which is the core requirement for sending a review link to a client.

**Fix:** set `FLOOM_SECRET` (the HMAC signing secret used by `routers/approvals.py` to mint/verify public share tokens) on the Railway cloud API env, then verify `/approvals/public/{id}?token=...` returns 200 and the owner share-link endpoint echoes a non-null `public_link`.

Verified live: created a real pending approval, owner-authenticated page renders fine; public token path 503s.

itachi-hue:
Fixed in cloud PR floomhq/workeros-cloud#533 (merged → auto-deploying to prod).

**RCA:** the engine code was already correct — `core/approval_signing.py` resolves the signer as `WORKEROS_APPROVAL_SIGNING_SECRET` → `FLOOM_SECRET` → None and fails closed (503) when none is set (#998). Cloud deliberately strips `FLOOM_SECRET` and never set the dedicated var, so the signer was None → every public approval/review link 503'd (the serializer mints a link per row, so `GET /api/approvals` 503'd the whole list).

**Fix:** cloud `startup.py` now derives a stable `WORKEROS_APPROVAL_SIGNING_SECRET` from the Supabase service-role key (HMAC-SHA256, domain-separated, explicit env wins), mirroring the existing `_ensure_magic_link_secret` helper. Code-reviewed by Codex (which caught + got fixed an alias-env edge case where `get_cloud_settings()` accepts `WORKEROS_CLOUD_SUPABASE_SERVICE_ROLE_KEY` but the helpers read only the bare name). Unit-tested; cloud CI green.

Issue #998, opened 2026-06-12
[security] multiple HMAC token signers fall back to hardcoded 'dev-secret-not-set'


federicodeponte:
## Problem
Several HMAC-based signing surfaces fall back to the public constant `"dev-secret-not-set"` when `FLOOM_SECRET` is not configured:

| Surface | Location |
|---|---|
| Worker-call token signing | `apps/api/run_token.py:110` (already #994/#972) |
| Webhook legacy token | `apps/api/webhook_service.py:66` |
| Worker public share token | `apps/api/main.py:7301` |
| Workspace template share token | `apps/api/main.py:11815` |
| Approval public token | `apps/api/main.py:13133`, `apps/api/chat_service.py:3765` |

Any deployment where `FLOOM_SECRET` is missing signs these tokens with a known string, allowing forgery.

## Impact
- Public share links for workers/brain files can be forged.
- Webhook legacy grace tokens can be forged.
- Approval review links can be forged.

## Suggested fix
1. Remove `"dev-secret-not-set"` fallbacks entirely.
2. For each signer, raise a clear error / return 503 when the effective secret is missing.
3. Require `FLOOM_SECRET` to be set in production; fail startup if absent.
4. Add tests covering each path refusing to operate without a real secret.

## References
- #972 (worker-call token fallback)
- `apps/api/webhook_service.py:66`
- `apps/api/main.py:7301`, `:11815`, `:13133`
- `apps/api/chat_service.py:3765`

itachi-hue:
Fixed in PR #1002. Removed the `dev-secret-not-set` fallback from every remaining signer: worker/workspace/approval public-share tokens (main.py, chat_service.py) now 503 without FLOOM_SECRET; the legacy webhook token (webhook_service.py) raises / fails closed in verify. Finishes the #972 sweep. Test: `tests/test_998_signer_fallback_removed.py` (literal gone everywhere; webhook fails closed; main signers 503).

PR #1747, opened 2026-06-21
fix(web): scroll standalone approval review page (#1717) + investigate share-link 503 (#1716)

federicodeponte:
## Summary

Fixes the approval review page scroll/clip bug (#1717) and reports the root cause for the public share-link 503 (#1716).

## #1717 — Approval review page doesn't scroll (CODE FIX)

**Root cause:** The standalone `AppShell` branch (matched by `/approvals/review` and the other `standalonePrefixes`) rendered `<main>` as `min-h-screen w-full` with **no scroll container**. The `<body>` is `h-screen overflow-hidden` (see `app/layout.tsx`), so any page content taller than the viewport grew past the fold with nothing able to scroll to it. A multi-item "Proposed output" (e.g. a 5-candidate shortlist rendered by `ApprovalActionItems`) was clipped after the first ~1.5 items, later items were unreachable, and a large empty area appeared below — exactly the reported symptom.

The in-app inbox path (`/approvals` → `ApprovalsCollection`) was already fine: it renders the review body inside `.c-dbody { flex:1; overflow:auto; }`. Standard app pages also already scroll via `StandardShellBody`'s `overflow-y-auto` pane. Only the standalone pane was missing its scroller.

**Fix:** Make the standalone `<main>` the scroll container with `h-full overflow-y-auto`, matching the established standard-page pattern. `h-full` resolves against the body's definite `h-screen` height. Short pages are unaffected (no scrollbar when content fits); tall pages now scroll on desktop and mobile. This also fixes the same latent clip for the other standalone routes (`/w`, `/s`, `/login`, `/run`, `/preview`, `/cli-auth`).

**Test:** Added `apps/web/tests/app-shell-standalone-scroll.dom.test.tsx` asserting the standalone main is an `overflow-y-auto` pane (and not `overflow-hidden`). Verified it FAILS on the pre-fix `min-h-screen` code and PASSES after.

## #1716 — Public share links 503 (DEPLOY/ENV FINDING, no code change)

Investigated `apps/api/main.py:auth_middleware`, `routers/approvals.py`, and `core/approval_signing.py`. **This is a deployment/env misconfiguration, not a code bug.** The code is already correct:

1. `/approvals/public/` is already in the auth-exempt allowlist (`main.py` ~L1671), so no-login signed links are not blocked by the operator-secret gate.
2. The signing secret is already decoupled from `FLOOM_SECRET` via the dedicated `WORKEROS_APPROVAL_SIGNING_SECRET` (with `FLOOM_SECRET` fallback) in `core/approval_signing.py::_approval_signing_secret()`. This is documented in `apps/api/.env.example` (L237-245).
3. The owner's `GET /approvals` already degrades gracefully (`try_approval_public_token` → `public_link: null`) instead of 503-ing the whole list.

The 503 on `GET /approvals/public/{id}?token=...` originates in `_load_public_approval` → `approval_public_token()`, which **fail-closes (503) by design** when NEITHER `WORKEROS_APPROVAL_SIGNING_SECRET` nor `FLOOM_SECRET` is set (#998: never mint/verify a share token against a public constant, or links become forgeable). Making this endpoint return 200 without a configured signer would be a security regression, so no code fix is appropriate.

**Exact config fix (Railway cloud API env):** set a stable secret value for `WORKEROS_APPROVAL_SIGNING_SECRET` (preferred for the multi-tenant host, since `startup.py` intentionally strips/blocks `FLOOM_SECRET`). After setting it and redeploying:
- `GET /approvals/public/{id}?token=...` returns 200,
- `GET /approvals` echoes a non-null `public_link`.

No code change is included for #1716.

## Test / build results (local)

- `apps/web`: `npm run lint` → 0 errors (pre-existing warnings only, none in changed files).
- `apps/web`: `npm run build` → Compiled successfully; `/approvals/review` route built.
- `apps/web`: new test file → 2 passed. Targeted AppShell suite → 4 passed.
- Full `vitest run` → 591 passed; 5 timeouts in unrelated `workers`/`collection` DOM tests caused by 5s default timeout under concurrent load (the full suite ran ~495s). Each flaked file passes in isolation in ~2s and is unrelated to this `AppShell` change.


federicodeponte (commented):
Adversarial review — verdict: APPROVE-WITH-NITS. The CSS height chain checks out (html `height:100%` → body `h-screen` definite → `<main>` `h-full`=100% resolves against body's definite height → `overflow-y-auto` scrolls). `<main>` is the sole in-flow flex child of body (BootSplash=null, Ambient/IconSprite are fixed/absolute/display:none, Toaster portals), so no flex-sibling contention. Old `min-h-screen` + body `overflow-hidden` genuinely clipped tall content — bug claim valid. #1716 analysis is correct: `/approvals/public/` is auth-exempt (main.py:1671), signer is decoupled via WORKEROS_APPROVAL_SIGNING_SECRET with FLOOM_SECRET fallback, and the 503 in `_load_public_approval`→`approval_public_token` is intentional fail-closed (#998: never verify against a public constant). Deploy-env conclusion stands; no code fix warranted.

Nit (test coverage gap) — apps/web/tests/app-shell-standalone-scroll.dom.test.tsx: the test only asserts `className` contains `overflow-y-auto` and not `overflow-hidden`. The actual bug required TWO things: a bounded height (`h-full`) AND the scroller (`overflow-y-auto`). A future edit that drops `h-full` while keeping `overflow-y-auto` would reintroduce the clip (height:auto → unbounded pane, nothing to scroll within) yet keep this test green. Suggest also asserting `expect(main?.className).toContain("h-full")` so the test guards the full fix, not just the overflow declaration. (Computed scroll behavior can't be asserted under jsdom, so the className proxy is the right approach — just widen it to the load-bearing class.)

Nit (PR description only, not code) — the body references `startup.py` for the FLOOM_SECRET strip/guard, but no `startup.py` exists in this repo; the strip lives in the downstream multi-tenant host (per `chore: remove hosted deploy residue from public tree`). The technical conclusion is unaffected, but the file reference is inaccurate.

No blocking issues. Not approving/merging per review scope.

itachi-hue:
Folded into and superseded by the integration PR #1749 (engine UI batch — review + merge as one). Closing to consolidate; the branch is preserved, so this can be reopened if #1749 drops it. Review/merge happens on #1749.

Issue #524, opened 2026-06-07
FL28: Verify full worker setup and Emily across UI/MCP/WhatsApp/Slack


federicodeponte:
## Description
Full worker-setup flow plus Emily across UI, MCP, WhatsApp, and Slack was not fully tested. Source says UI and WhatsApp were done; Slack and MCP were pending.

## Evidence
Source: /tmp/fede-livetest-issues-2026-06-07.md, FL28.
PR: #495 documents MCP and Slack verification.

## App
OSS and integrations.

## Expected
All four channels have live verification evidence, and any channel-specific defects are filed separately.

itachi-hue:
This is a QA verification task, not a code change. It requires a human to manually run through the worker setup flow on each channel (UI, MCP, WhatsApp, Slack) and record evidence. No code can be written to close this. Skipping this batch; needs a live test session.

itachi-hue:
Skipping this batch. This is a full end-to-end QA pass (worker setup, Emily across UI/MCP/WhatsApp/Slack) — requires a live staging environment and manual testing. Cannot be resolved by a code change.

federicodeponte:
## OSS MCP verification added (2026-06-08)

**HTTP MCP (install path):** `POST https://workers-api.floom.dev/mcp-tools/serve` with `x-floom-secret`
- initialize ✅
- tools/list ✅
- tools/call `workers.list` ✅

**stdio `workeros-mcp`:** 58 tools; prod probes OK for workers/runs/connections/contexts/approvals + `runs.watch` (AX41 Node 22, direct `node dist/server.js`).

Still **not** re-verified this pass: Slack/WhatsApp MCP bridges, Cloud MCP (`/mcp/{workspace}`).

Related: #617 demo checklist, new docs/packaging issues for README vs HTTP install.

federicodeponte:
## Investigation (read-only, 2026-06-09)

**Status: mostly verified; close when PR #495 lands.**

PR [#495](https://github.com/floomhq/workeros/pull/495) (OPEN) documents live matrix (2026-06-07):
| Channel | Evidence |
|---------|----------|
| UI web chat | Prior sessions ✅ |
| WhatsApp | Prior sessions ✅ |
| MCP stdio | `workspace.chat` round-trip ✅ |
| Slack | Signed `app_mention` → 200 queued → Emily reply ✅ |

**Remaining gaps (file separately or extend #495):**
[...]

PR #1757, opened 2026-06-21
Fix issue #524

itachi-hue:
Autonomous WorkerOS v1 fix for https://github.com/floomhq/workeros/issues/524.


itachi-hue:
Closing this spike PR: the first autonomous run proved end-to-end PR creation, but the generated change overclaimed verification and the worker has since been tightened to use structured file edits plus actual test-result reporting.

Issue #517, opened 2026-06-07
FL21: Brain file viewer needs much more reading space


federicodeponte:
## Description
Opening a Brain file does not provide enough reading space by default.

## Evidence
Source: /tmp/fede-livetest-issues-2026-06-07.md, FL21.

## App
OSS.

## Expected
File opening prioritizes a large reading viewer by default.

itachi-hue:
Fixed in commit 64b8fe5 on branch `fix/fl-batch-2-ui-issues`.

**Root cause:** `MID_MAX` (the maximum width of the file viewer pane) was capped at 560px and `MID_DEFAULT` was 280px — both too narrow for comfortable file reading, especially markdown or code files.

**Fix:** `apps/web/app/contexts/page.tsx` — `MID_MAX` 560 → 900, `MID_DEFAULT` 280 → 360. The viewer now opens wider by default and can be dragged much further before hitting the cap.

**Unit test:** `apps/web/tests/fl-batch-2.test.ts` — `#517 Brain pane widths`
Reads `contexts/page.tsx`, extracts the `MID_MAX` and `MID_DEFAULT` constants, and asserts `MID_MAX > 560` and `MID_DEFAULT ≤ MID_MAX`. Test passes (5/5).

PR #1762, opened 2026-06-21
fix: enforce result.json in generated run.py

itachi-hue:
## Summary
- add a pre-persist generated-run.py contract guard for worker-author/smoke-repair output
- reject LLM-generated scripts that write legacy `outputs.json` or `output.json`
- reject generated scripts that never mention/write `result.json`
- strengthen the repair prompt to explicitly ban legacy output filenames

Fixes cloud issue floomhq/workeros-cloud#517 once cloud bumps the engine submodule.

## Tests
- `python -m pytest apps/api/tests/test_1416_worker_authoring_contract_docs.py apps/api/tests/test_517_emily_result_json_contract.py -q`
- `python -m py_compile apps/api/services/run_authoring.py apps/api/tests/test_517_emily_result_json_contract.py`


itachi-hue (commented):
**Blocking review, posted as COMMENT because GitHub rejected `REQUEST_CHANGES` from the PR author identity.**

PR class: `code_fix`

The PR introduces a pre-persist contract guard to prevent LLM-generated scripts from writing legacy output filenames (`outputs.json`, `output.json`) and to ensure they write `result.json`. While the goal is correct, the implementation uses naive substring checks on the entire code string, which will cause false positives and discard valid code if the LLM includes comments or docstrings mentioning the legacy filenames (which is highly likely given the updated prompt instructions).

Findings:
- **medium**: The substring check `name in lowered` in `_generated_run_py_contract_error` will trigger false positives if the LLM includes comments or docstrings referencing the banned legacy filenames (e.g., `# Do not write outputs.json`). Since the prompt explicitly instructs the LLM to 'NEVER write outputs.json or output.json', the LLM is highly likely to include these filenames in comments or explanations, causing valid repaired code to be discarded. Consider stripping comments/docstrings or parsing the AST to check for actual string literals or file-writing operations.
- **low**: The check `"result.json" not in lowered` is also a simple substring check. It can be bypassed by a comment like `# Will write result.json` without actually writing the file, or conversely, it could fail if the code writes to a variable or uses a dynamic path construction that doesn't literally contain the string 'result.json'.

Reviewer: WorkerOS Gemini PR Reviewer V1.

itachi-hue:
Updated after review. The result.json contract guard now runs in the initial registration path before appending generated run.py, so invalid first-pass code does not reach _register_worker_from_files and the drafted bundle remains viewable. Added registration-path tests for outputs.json, output.json, missing result.json, and valid result.json, plus kept the repair-path coverage.\n\nVerification:\n- python -m pytest apps/api/tests/test_517_emily_result_json_contract.py -q -> 6 passed\n- python -m pytest apps/api/tests/test_worker_author_registration_smoke_guard.py apps/api/tests/test_1416_worker_authoring_contract_docs.py apps/api/tests/test_517_emily_result_json_contract.py -q -> 10 passed\n- python -m py_compile apps/api/services/run_authoring.py apps/api/tests/test_517_emily_result_json_contract.py -> passed

PR #1791, opened 2026-06-21
ci: deploy staging web from self-hosted runner

itachi-hue:
Adds a staging deploy workflow that runs only on the self-hosted runner.\n\n- Triggers on pushes to staging and workflow_dispatch\n- Uses Vercel CLI, not GitHub-hosted Actions minutes\n- Deploys the web app from apps/web\n- Aliases staging to workeros-staging.vercel.app\n- Points the frontend at the existing prod API backend


itachi-hue:
Closing/defer. This staging Vercel workflow is optional ops work, still references the old workeros staging target, and is not part of the current OSS launch/code path. It should be reopened as a fresh ops PR if we decide to own a self-hosted staging deploy flow.

Issue #1787, opened 2026-06-21
fix(secrets): scrub_secrets blind-replaces short/low-entropy secret values → mass false redaction of legitimate output


itachi-hue:
## Symptom
A worker's output had a candidate title **"Full Stack Engineer"** render as:
```
<REDACTED:EXTERNAL_APIFY_PROFILE_SCRAPER_MODE> Stack Engineer
```
The secret `EXTERNAL_APIFY_PROFILE_SCRAPER_MODE` has the value `Full`, and the scrubber blind-replaced **every** occurrence of the string `Full` in the output. Any declared secret whose value is a short/common word silently corrupts unrelated legitimate text (titles, names, descriptions, logs).

## Root cause
`apps/api/run_service.py` `scrub_secrets` (and `scrub_secret_values` for persisted outputs):
```python
for name, value in secrets.items():
    if value and len(value) > 3:
        text = text.replace(value, f"<REDACTED:{name}>")
```
The only gate is `len(value) > 3`, so `Full`, `auto`, `true`, `both`, `Java`, region codes, model names, etc. all qualify and get globally substring-replaced. It's a raw `str.replace` (no word boundary, no entropy check, no secret-class awareness).

## The constraint (do NOT regress security)
This must reduce **false positives** without creating a path for a **real** secret to leak. Safe-by-default: anything credential-shaped stays redacted. We only stop redacting values that are *provably* not secrets.

## Proposed design (tiered, security-first)
Decide redaction per secret by value quality + name class, not a flat length:
1. **Always redact** (unchanged security):
   - Values matching the existing `SECRET_PATTERNS` (`sk_/pk_ ...`, `key|secret|token|password = ...`), JWTs, and any long high-entropy token.
   - Any secret whose **name** is credential-class (`*KEY*`, `*TOKEN*`, `*SECRET*`, `*PASSWORD*`, `*CRED*`, `*_JSON` cred blobs) — regardless of value length.
2. **Conditionally redact** short/low-entropy values: only on **word boundaries** (`\bvalue\b`), and skip when the value is a common dictionary word / all-lowercase-alpha token AND the secret name is config-class (`*MODE*`, `*REGION*`, `*CONCURRENCY*`, `*BATCH*`, `*THRESHOLD*`, `*MODEL*`, `*_PATH`, `*_URL` non-credential, numeric caps).
[...]

PR #1792, opened 2026-06-21
Fix issue #1787 via Claude/Codex loop

itachi-hue:
Automated WorkerOS fix loop for #1787.

VPS workspace: /home/dev/code_dir/workerOS-refactor
Branch: workeros/claude-codex-issue-1787-1782074023


itachi-hue:
Automated codex review iteration 1

```text
+++ b/apps/api/run_service.py
@@ -5,6 +5,7 @@ import uuid
 import json
 import threading
 import re
+import math
 import logging
 import shutil
 import time
@@ -549,16 +550,99 @@ def publish_run_part(run_id: str, part: dict) -> None:
 SECRET_PATTERNS = [
     re.compile(r"(?i)(api[_-]?key|secret|token|password)\s*[=:]\s*[^\s'\"]+"),
     re.compile(r"\b(?:sk|pk)_(?:live|test|proj|sec)_[a-zA-Z0-9_-]+\b"),
+    # JWTs: three base64url segments separated by dots.
+    re.compile(r"\beyJ[a-zA-Z0-9_-]+\.[a-zA-Z0-9_-]+\.[a-zA-Z0-9_-]+\b"),
 ]
 
+# Secret *names* that are credential-class: their values are always redacted
+# (raw substring replace) regardless of value length or entropy.
+_CREDENTIAL_NAME_RE = re.compile(
+    r"(?i)(KEY|TOKEN|SECRET|PASSWORD|PASSWD|CRED|AUTH|PRIVATE|PIN|"
+    r"SIGNATURE|SIGNING|CERT|SESSION|COOKIE|BEARER|_JSON$)"
+)
+
+# Secret *names* that are config-class: short/low-entropy values declared under
+# these are mis-declared config flags, not secrets (the "Full"/MODE case).
+_CONFIG_NAME_RE = re.compile(
+    r"(?i)(MODE|REGION|CONCURRENCY|BATCH|THRESHOLD|MODEL|LIMIT|TIMEOUT|"
+    r"RETRY|RETRIES|COUNT|SIZE|INTERVAL|LEVEL|FORMAT|LOCALE|LANG|VERSION|"
+    r"_PATH$|_URL$)"
+)
+
+# A short, single, common-word-shaped value (pure alpha, no entropy). These get
+# globally substring-replaced under the old code, mangling unrelated text.
+_LOW_ENTROPY_WORD_RE = re.compile(r"[A-Za-z]+")
+
+
+def _shannon_entropy(value: str) -> float:
+    counts: Dict[str, int] = {}
+    for char in value:
+        counts[char] = counts.get(char, 0) + 1
+    length = len(value)
+    entropy = 0.0
+    for count in counts.values():
+        p = count / length
+        entropy -= p * math.log2(p)
+    return entropy
+
+
+def _value_is_credential_shaped(value: str) -> bool:
+    """True when the value itself looks like a real credential and must be
+    redacted even if its declared name is not credential-class."""
+    if len(value) >= 20:
+        return True
+    for pattern in SECRET_PATTERNS:
+        if pattern.search(value):
+            return True
+    has_alpha = any(c.isalpha() for c in value)
+    has_digit = any(c.isdigit() for c in value)
+    has_symbol = any(not c.isalnum() for c in value)
+    # Mixed character classes at a credential-ish length.
+    if len(value) >= 8 and ((has_alpha and has_digit) or has_symbol):
+        return True
+    # Long, high-entropy single-class tokens (e.g. 16-char alpha key).
+    if len(value) >= 16 and _shannon_entropy(value) >= 3.0:
+        return True
+    return False
+
+
+def _is_low_entropy_word(value: str) -> bool:
+    """True for a short, plain word/number unlikely to be a secret."""
+    return len(value) < 8 and bool(_LOW_ENTROPY_WORD_RE.fullmatch(value))
+
 
 def scrub_secrets(text: str, secrets: Dict[str, str]) -> str:
-    """Replace secret values with redacted markers in log messages."""
+    """Replace secret values with redacted markers in log messages.
+
+    Tiered, security-first redaction: credential-named or credential-shaped
+    values are always redacted; provably non-secret short/low-entropy values
+    declared under config-class names are left intact (they are mis-declared
+    config flags, and blind substring-replacing them corrupts legitimate
+    output such as titles and descriptions).
+    """
     if not text:
         return text
     for name, value in secrets.items():
-        if value and len(value) > 3:
+        if not value or len(value) <= 3:
+            continue
+        # Tier 1: always redact (security unchanged for real secrets).
+        if _CREDENTIAL_NAME_RE.search(name) or _value_is_credential_shaped(value):
             text = text.replace(value, f"<REDACTED:{name}>")
+            continue
+        # Tier 3: never blind-replace a short common word under a config-class
+        # name. Warn so operators move it to env: instead of secrets:.
+        if _CONFIG_NAME_RE.search(name) and _is_low_entropy_word(value):
+            logger.warning(
+                "Declared secret %s has a low-entropy/config-shaped value; "
+                "consider declaring it in env: not secrets:",
+                name,
+            )
+            continue
+        # Tier 2: redact short/ambiguous values, but only on word boundaries so
+        # they cannot mangle unrelated substrings of legitimate output.
+        text = re.sub(
+            rf"\b{re.escape(value)}\b", f"<REDACTED:{name}>", text
+        )
     for pattern in SECRET_PATTERNS:
         text = pattern.sub("<REDACTED>", text)
     return text
diff --git a/apps/api/tests/test_1787_scrub_secrets_low_entropy.py b/apps/api/tests/test_1787_scrub_secrets_low_entropy.py
new file mode 100644
index 00000000..d688a3f9
--- /dev/null
+++ b/apps/api/tests/test_1787_scrub_secrets_low_entropy.py
@@ -0,0 +1,81 @@
+from __future__ import annotations
+
+import logging
+import sys
+from pathlib import Path
+
+API_DIR = Path(__file__).resolve().parents[1]
+if str(API_DIR) not in sys.path:
+    sys.path.insert(0, str(API_DIR))
+
+import run_service
+from run_service import scrub_secrets, scrub_secret_values
+
+
+def test_short_config_word_not_blind_replaced(caplog):
+    """The reported bug: a config flag declared as a secret with value 'Full'
+    must not redact every 'Full' in legitimate output."""
+    secrets = {"EXTERNAL_APIFY_PROFILE_SCRAPER_MODE": "Full"}
+    with caplog.at_level(logging.WARNING, logger="floom.run_service"):
+        result = scrub_secrets("Full Stack Engineer", secrets)
+    assert result == "Full Stack Engineer"
+    # Operator gets warned to move it to env:.
+    assert any("low-entropy" in rec.message for rec in caplog.records)
+
+
+def test_real_api_key_value_still_redacted():
+    secrets = {"APIFY_API_KEY": "a1b2c3d4e5f6g7h8i9j0k1l2"}
+    result = scrub_secrets("token is a1b2c3d4e5f6g7h8i9j0k1l2 here", secrets)
+    assert "a1b2c3d4e5f6g7h8i9j0k1l2" not in result
+    assert "<REDACTED:APIFY_API_KEY>" in result
+
+
+def test_credential_named_short_value_still_redacted():
+    secrets = {"API_PIN": "1234"}
+    result = scrub_secrets("the pin is 1234 ok", secrets)
+    assert "<REDACTED:API_PIN>" in result
+    assert "1234" not in result
+
+
+def test_sk_live_and_jwt_patterns_still_redacted():
+    text = (
+        "key sk_live_abcDEF123 and "
+        "jwt eyJhbGciOiJIUzI1NiJ9.eyJzdWIiOiIxIn0.abcDEF-_123"
+    )
+    result = scrub_secrets(text, {})
+    assert "sk_live_abcDEF123" not in result
+    assert "eyJhbGciOiJIUzI1NiJ9" not in result
+    assert result.count("<REDACTED>") >= 2
+
+
+def test_high_entropy_value_redacted_even_with_config_name():
+    # Config-class name but value is clearly credential-shaped -> still redacted.
+    secrets = {"MODEL_REGION": "k7Qz9_aB3xR-2mN4pLs8"}
+    result = scrub_secrets("region=k7Qz9_aB3xR-2mN4pLs8", secrets)
+    assert "k7Qz9_aB3xR-2mN4pLs8" not in result
+    assert "<REDACTED:MODEL_REGION>" in result
+
+
+def test_word_boundary_value_does_not_mangle_substrings():
+    # Non-credential, non-config name, ambiguous value -> word-boundary only.
+    secrets = {"SCRAPER_TAG": "auto"}
+    result = scrub_secrets("automatic auto pilot", secrets)
+    assert result == "automatic <REDACTED:SCRAPER_TAG> pilot"
+
+
+def test_persisted_output_dict_value_quality_gate():
+    secrets = {
+        "EXTERNAL_APIFY_PROFILE_SCRAPER_MODE": "Full",
+        "OUTPUT_SECRET": "sk-legacy-output-secret-9af",
+    }
+    output = {
+        "title": "Full Stack Engineer",
+        "blob": "prefix-sk-legacy-output-secret-9af",
+        "items": ["Full time", {"v": "sk-legacy-output-secret-9af"}],
+    }
+    safe = scrub_secret_values(output, secrets)
+    assert safe["title"] == "Full Stack Engineer"
+    assert safe["items"][0] == "Full time"
+    assert "sk-legacy-output-secret-9af" not in safe["blob"]
+    assert "<REDACTED:OUTPUT_SECRET>" in safe["blob"]
+    assert safe["items"][1]["v"] == "<REDACTED:OUTPUT_SECRET>"

exec
/bin/bash -lc "grep -R \"def scrub_secret_values\" -n apps/api/run_service.py && sed -n '520,680p' apps/api/run_service.py && grep -R \"scrub_secrets\" -n apps/api | head -50" in /home/dev/code_dir/workerOS-refactor
 succeeded in 8542ms:
grep: apps/api/tests/__pycache__/test_1787_scrub_secrets_low_entropy.cpython-314-pytest-9.0.3.pyc: binary file matches
grep: apps/api/services/__pycache__/ai_observability.cpython-314.pyc: binary file matches
grep: apps/api/__pycache__/run_service.cpython-314.pyc: binary file matches
apps/api/tests/test_1787_scrub_secrets_low_entropy.py:12:from run_service import scrub_secrets, scrub_secret_values
apps/api/tests/test_1787_scrub_secrets_low_entropy.py:20:        result = scrub_secrets("Full Stack Engineer", secrets)
apps/api/tests/test_1787_scrub_secrets_low_entropy.py:28:    result = scrub_secrets("token is a1b2c3d4e5f6g7h8i9j0k1l2 here", secrets)
apps/api/tests/test_1787_scrub_secrets_low_entropy.py:35:    result = scrub_secrets("the pin is 1234 ok", secrets)
apps/api/tests/test_1787_scrub_secrets_low_entropy.py:45:    result = scrub_secrets(text, {})
apps/api/tests/test_1787_scrub_secrets_low_entropy.py:54:    result = scrub_secrets("region=k7Qz9_aB3xR-2mN4pLs8", secrets)
apps/api/tests/test_1787_scrub_secrets_low_entropy.py:62:    result = scrub_secrets("automatic auto pilot", secrets)
apps/api/services/ai_observability.py:750:        exc_value = _redact_pii(_scrub_secrets(str(exc)))
apps/api/services/ai_observability.py:754:        stack = _redact_pii(_scrub_secrets(stack))
apps/api/services/ai_observability.py:847:    scrubbed = _scrub_secrets(text)
apps/api/services/ai_observability.py:868:def _scrub_secrets(text: str) -> str:
apps/api/run_service.py:614:def scrub_secrets(text: str, secrets: Dict[str, str]) -> str:
apps/api/run_service.py:654:        return scrub_secrets(value, secrets)
apps/api/run_service.py:2720:        safe_msg = scrub_secrets(msg, run_secrets)

exec
/bin/bash -lc "nl -ba apps/api/run_service.py | sed -n '632,648p'" in /home/dev/code_dir/workerOS-refactor
 succeeded in 0ms:
   632	        # Tier 3: never blind-replace a short common word under a config-class
   633	        # name. Warn so operators move it to env: instead of secrets:.
   634	        if _CONFIG_NAME_RE.search(name) and _is_low_entropy_word(value):
   635	            logger.warning(
   636	                "Declared secret %s has a low-entropy/config-shaped value; "
   637	                "consider declaring it in env: not secrets:",
   638	                name,
   639	            )
   640	            continue
   641	        # Tier 2: redact short/ambiguous values, but only on word boundaries so
   642	        # they cannot mangle unrelated substrings of legitimate output.
   643	        text = re.sub(
   644	            rf"\b{re.escape(value)}\b", f"<REDACTED:{name}>", text
   645	        )
   646	    for pattern in SECRET_PATTERNS:
   647	        text = pattern.sub("<REDACTED>", text)
   648	    return text

codex
The patch improves the low-entropy config case, but it introduces a redaction gap for short configured secrets with punctuation boundaries. That can leak secrets in logs or persisted outputs, so the patch should not be considered fully correct.

Full review comments:

- [P2] Redact short secrets that start or end with punctuation — /home/dev/code_dir/workerOS-refactor/apps/api/run_service.py:643-644
  For secrets with non-credential names and short values that begin or end with a non-word character, this `\b...\b` pattern does not match even when the value appears standalone, e.g. `scrub_secrets("salt $abc$", {"SALT": "$abc$"})` leaves `$abc$` in the output. These values were previously redacted by raw replacement, so this change can leak configured secrets in logs or persisted output for that class of inputs.

- [P3] Treat numeric config values as low-entropy too — /home/dev/code_dir/workerOS-refactor/apps/api/run_service.py:594-594
  The helper is documented as covering short words/numbers, but `_LOW_ENTROPY_WORD_RE` only matches alphabetic values. As a result, config-class values such as `TIMEOUT=3000` or `LIMIT=1000` still fall through to redaction and can corrupt legitimate standalone numeric output, which is the same class of mis-declared config value this change is trying to avoid.
The patch improves the low-entropy config case, but it introduces a redaction gap for short configured secrets with punctuation boundaries. That can leak secrets in logs or persisted outputs, so the patch should not be considered fully correct.

Full review comments:

- [P2] Redact short secrets that start or end with punctuation — /home/dev/code_dir/workerOS-refactor/apps/api/run_service.py:643-644
  For secrets with non-credential names and short values that begin or end with a non-word character, this `\b...\b` pattern does not match even when the value appears standalone, e.g. `scrub_secrets("salt $abc$", {"SALT": "$abc$"})` leaves `$abc$` in the output. These values were previously redacted by raw replacement, so this change can leak configured secrets in logs or persisted output for that class of inputs.

- [P3] Treat numeric config values as low-entropy too — /home/dev/code_dir/workerOS-refactor/apps/api/run_service.py:594-594
  The helper is documented as covering short words/numbers, but `_LOW_ENTROPY_WORD_RE` only matches alphabetic values. As a result, config-class values such as `TIMEOUT=3000` or `LIMIT=1000` still fall through to redaction and can corrupt legitimate standalone numeric output, which is the same class of mis-declared config value this change is trying to avoid.
```

itachi-hue:
Closing after triage. The linked issue (#1787) was already closed as not an engine-side bug, and the proposed redaction change had an unresolved Codex blocking/security finding around punctuation-boundary short secrets. This should not be merged.

Issue #1797, opened 2026-06-21
[EPIC][P0] Cloud onboarding is end-to-end broken — a user cannot get from `login` to a deployed worker


federicodeponte:
Tracking epic. Over one real session trying to stand up a new workspace + deploy 5 pre-validated workers from a headless host, the cloud onboarding path failed at **every** step. Each link is filed separately; this epic captures the end-to-end break and the order it hits you.

## The failure chain (what actually happened, in order)

1. **Login won't complete reliably** — `login --cloud` on a headless host left orphaned login processes that re-served a **stale/consumed device code**; approving it gave "user code not found." Killing the zombies was required to get a fresh code. → #1740
2. **Approval page dead-ends** — opening the fresh `/app/cli-auth?code=...` link showed a bare **"unauthorized"** (web session not recognized) with no redirect-to-login, so the device code never got approved. → #1789
3. **No non-interactive fallback** — there is no API-key/token auth for cloud; the only path is the browser device flow above, so when it breaks you're fully stuck. OSS mode can't send `x-floom-user` either. → #1742
4. **Even when login briefly works, `workspaces create` kills the session** — create succeeds server-side but immediately invalidates the token, so you can create a workspace but can't push a single worker into it. → #1687
5. **Retrying create silently duplicates** — because step 4 made create *look* failed, retries produced **two `fede-secretary` and two `content-pipeline`** in one account, with no way to delete the orphans. → #1738
6. **Brain packs can't be set up headlessly** — there's no `workeros contexts` CLI, so a worker that needs a context/brain-pack can't be provisioned from a script/cron. → #1741

## Net effect

Build was 100% ready (5 `workers validate`-clean bundles + a CRM brain pack). **Zero** got deployed — not for any build reason, purely because the auth/workspace/onboarding surface fails at each step. The only working route discovered was bypassing the CLI entirely via an already-authed MCP session on another machine.

## Definition of done for this epic

A fresh user on a clean host can: `login --cloud` (or a token), create/select a workspace (no dupes, no session death), set up a connection + brain pack, and `workers push` + run — without hitting any of the above. Sub-issues: #1687, #1738, #1740, #1741, #1742, #1789. Related: #1666.

itachi-hue:
Claude/Codex automated loop did not merge.

Status: no_changes
Verdict: BLOCKED
Summary:
Claude completed but produced no working-tree changes.

itachi-hue:
[...]

PR #1798, opened 2026-06-21
Fix issue #1797 via Claude/Codex loop

itachi-hue:
Automated WorkerOS fix loop for #1797.

VPS workspace: /home/dev/code_dir/workerOS
Branch: workeros/claude-codex-issue-1797-1782076421


itachi-hue:
Automated codex review iteration 1

```text
def _context_detail(
    name: str,
    metadata: dict[str, dict[str, Any]] | None = None,
    *,
    repos: Optional[Repositories] = None,
    user_id: str = "local-user",
    path_prefix: str | None = None,
) -> ContextDetail:
    from contexts import context_dir, context_file_metadata, context_mount_names, iter_context_files, load_context_metadata
    from models import ContextDetail, ContextFileItem, ContextWorkerRef

    root = context_dir(name)
    if not root.is_dir():
        raise HTTPException(status_code=404, detail="Context not found")
    meta = metadata if metadata is not None else load_context_metadata()
    # #783: optional path_prefix narrows the (otherwise flat) file list to one
    # subfolder so the Brain UI can navigate nested folders. Normalized to a
    # trailing-slash dir prefix; matched against each file's posix relpath.
    norm_prefix = ""
    if path_prefix and path_prefix.strip("/"):
        norm_prefix = path_prefix.strip("/") + "/"
    files = [
        ContextFileItem(**context_file_metadata(root, path, pack_metadata=meta.get(name) or {}))
        for path in sorted(iter_context_files(root), key=lambda p: p.relative_to(root).as_posix())
        if not norm_prefix or path.relative_to(root).as_posix().startswith(norm_prefix)
    ]
    summary = _context_summary(name, meta, repos=repos, user_id=user_id)
    # Compute used_by and worker_count when repos is available.
    used_by: List[ContextWorkerRef] = []
    if repos is not None:
        try:
            for worker in repos.workers.list(user_id=user_id):
                try:
                    contexts = (worker.get("config") or {}).get("contexts") or []
                    if name in context_mount_names(contexts):
                        used_by.append(ContextWorkerRef(
                            worker_id=str(worker["id"]),
                            worker_name=str(worker.get("name") or worker["id"]),
                        ))
                except Exception:
                    continue
        except Exception:
            pass
    description = _context_description(root)
    if description is None and summary.system:
        description = _system_context_description(name)
    summary.worker_count = len(used_by)
    summary.description = description
    return ContextDetail(
        **summary.model_dump(),
        files=files,
        used_by=used_by,
    )


def _raise_context_quota_if_needed(name: str) -> None:
    from contexts import context_dir, context_total_size

    total = context_total_size(context_dir(name))
    limit = _context_total_limit_bytes()
    if total > limit:
        raise HTTPException(
            status_code=400,
            detail=f"Context exceeds {_format_bytes(limit)} total size limit ({total} bytes)",

exec
/bin/bash -lc "sed -n '120,160p' apps/api/services/context_access.py && sed -n '340,390p' apps/api/contexts.py && sed -n '1,70p' apps/api/tests/test_1052_path_validator_percent_encoding.py" in /home/dev/code_dir/workerOS
 succeeded in 0ms:
        return validate_context_name(name)
    except ValueError as exc:
        raise HTTPException(status_code=400, detail=str(exc)) from exc


def _context_file_path_or_400(path: str) -> str:
    from contexts import normalize_context_file_path
    try:
        return normalize_context_file_path(path)
    except ValueError as exc:
        raise HTTPException(status_code=400, detail=str(exc)) from exc


def _safe_context_file_or_400(name: str, path: str) -> Path:
    from contexts import safe_context_file_path
    try:
        return safe_context_file_path(name, path)
    except ValueError as exc:
        raise HTTPException(status_code=400, detail=str(exc)) from exc


def _unowned_contexts_visible_to_caller() -> bool:
    return os.environ.get("WORKEROS_ENABLE_USER_HEADER_SCOPE") != "1" and not _is_cloud_deploy()


def _is_system_context_pack(
    name: str,
    metadata: dict[str, dict[str, Any]] | None = None,
) -> bool:
    """True for engine/system packs that must be hidden from operators."""
    from contexts import load_context_metadata, validate_context_name
    try:
        safe_name = validate_context_name(name)
    except ValueError:
        return False
    if safe_name in SYSTEM_CONTEXT_PACKS:
        return True
    meta = metadata if metadata is not None else load_context_metadata()
    return bool((meta.get(safe_name) or {}).get("system"))


        raise ValueError(f"Path traversal attempt: {target}")
    # Cloud lazy-hydration: if the directory is absent or empty and a hydration
    # hook is registered (cloud startup sets one), fetch the context pack from
    # Supabase Storage before returning the path.  This is a no-op in OSS /
    # single-tenant mode (_hydration_hook is None) and a no-op whenever the
    # directory is already populated (idempotency guard is inside the hook).
    hook = _hydration_hook
    if hook is not None and (not target.exists() or not any(target.iterdir())):
        try:
            hook(_current_scope(), safe_name, target)
        except Exception:
            pass  # hydration failures must never block a run
    return target


def safe_context_file_path(name: str, raw_path: str) -> Path:
    rel = normalize_context_file_path(raw_path)
    root = context_dir(name)
    target = (root / rel).resolve()
    if not _is_relative_to_path(target, root):
        raise ValueError(f"Path traversal attempt: {target}")
    return target


def normalize_context_file_path(raw_path: str) -> str:
    # #1052 — reject percent-encoded path separators (%2f, %5c) so this validator
    # and the worker-file validator agree; they were previously accepted as
    # literal filenames.
    lowered = (raw_path or "").lower()
    if "%2f" in lowered or "%5c" in lowered:
        raise ValueError(f"invalid context file path: {raw_path!r}")
    # #1077 — reject NUL and other ASCII control characters. A %00 in the URL
    # decodes to a literal null byte that later reached Path.is_file()/lstat and
    # raised an uncaught "embedded null character in path" (raw OS error -> 500).
    # Reject it here so every context-file route returns a clean 400 instead.
    if any(ord(ch) < 0x20 for ch in (raw_path or "")):
        raise ValueError(f"invalid context file path: {raw_path!r}")
    text = (raw_path or "").replace("\\", "/").strip("/")
    path = PurePosixPath(text)
    if not text or str(path) == ".":
        raise ValueError("file path is required")
    if path.is_absolute() or any(part in {"", ".", ".."} for part in path.parts):
        raise ValueError(f"invalid context file path: {raw_path!r}")
    return path.as_posix()


def ensure_contexts_dir() -> None:
    current_contexts_root().mkdir(parents=True, exist_ok=True)


def load_context_metadata() -> dict[str, dict[str, Any]]:
"""#1052 (item 2) — path-validator inconsistency: `..%2f` was accepted as a
literal filename by workers.write_file while contexts.write rejected the decoded
traversal. Both validators now reject percent-encoded path separators (%2f/%5c)
so they agree.
"""
from __future__ import annotations

import sys
from pathlib import Path

import pytest
from fastapi import HTTPException

API_DIR = Path(__file__).resolve().parents[1]
if str(API_DIR) not in sys.path:
    sys.path.insert(0, str(API_DIR))

import main
from contexts import normalize_context_file_path

ENCODED = ["..%2fzz.txt", "..%2Fzz.txt", "..%5cx", "..%5Cx", "dir%2f..%2fetc"]


@pytest.mark.parametrize("bad", ENCODED)
def test_worker_file_path_rejects_percent_encoded_separators(bad):
    with pytest.raises(HTTPException) as exc:
        main._validate_worker_file_path(bad)
    assert exc.value.status_code == 400


@pytest.mark.parametrize("bad", ENCODED)
def test_context_file_path_rejects_percent_encoded_separators(bad):
    with pytest.raises(ValueError):
        normalize_context_file_path(bad)


def test_worker_file_path_accepts_normal_names():
    for ok in ("run.py", "lib/helper.py", "docs/notes.md"):
        main._validate_worker_file_path(ok)  # no raise


def test_context_file_path_accepts_normal_names():
    assert normalize_context_file_path("notes.md") == "notes.md"
    assert normalize_context_file_path("a/b/c.txt") == "a/b/c.txt"


def test_existing_plain_traversal_still_rejected():
    with pytest.raises(HTTPException):
        main._validate_worker_file_path("../etc/passwd")
    with pytest.raises(ValueError):
        normalize_context_file_path("../etc/passwd")

exec
/bin/bash -lc "nl -ba apps/mcp/src/commands/contexts.ts | sed -n '175,205p'" in /home/dev/code_dir/workerOS
 succeeded in 0ms:
   175	  }
   176	  try {
   177	    const bytes = await readFile(localFile);
   178	    const { client } = await createAuthenticatedClient();
   179	    // The endpoint writes the raw request body verbatim when the content-type
   180	    // is not application/json, so bytes round-trip for both text and binary.
   181	    const encodedPath = cleanPath.split("/").map(encodeURIComponent).join("/");
   182	    const result = (await client.requestJson(
   183	      "PUT",
   184	      `/contexts/${encodeURIComponent(name)}/files/${encodedPath}`,
   185	      { rawBody: new Uint8Array(bytes) },
   186	    )) as ContextFileItem;
   187	    if (options.json) {
   188	      printJson(result);
   189	      return 0;
   190	    }
   191	    log.ok(`Pushed ${cleanPath} to context ${name} (${result.size} bytes).`);
   192	    return 0;
   193	  } catch (error) {
   194	    if (error instanceof FloomApiError && error.status === 404) {
   195	      notFoundHint(name);
   196	      return 1;
   197	    }
   198	    if (error && typeof error === "object" && "code" in error && (error as { code?: string }).code === "ENOENT") {
   199	      log.err(`Local file not found: ${localFile}`);
   200	      return 1;
   201	    }
   202	    const handled = handleAuthError(error);
   203	    if (handled !== null) return handled;
   204	    throw error;
   205	  }

codex
The new CLI flow generally works, but it drops security warnings that the API explicitly returns for context writes. This can cause users to miss leaked secrets in uploaded packs.

Review comment:

- [P2] Surface secret warnings from context uploads — /home/dev/code_dir/workerOS/apps/mcp/src/commands/contexts.ts:191-191
  When a pushed file contains a high-confidence secret, the `/contexts/{name}/files/...` endpoint returns `has_secret_warning` and `secret_warnings` so operators can move the credential out of the brain pack, but the non-JSON CLI path prints only a success message. In the default warn mode this means `floom contexts push` succeeds while hiding the only warning about the leaked credential unless the user happened to pass `--json`.
The new CLI flow generally works, but it drops security warnings that the API explicitly returns for context writes. This can cause users to miss leaked secrets in uploaded packs.

Review comment:

- [P2] Surface secret warnings from context uploads — /home/dev/code_dir/workerOS/apps/mcp/src/commands/contexts.ts:191-191
  When a pushed file contains a high-confidence secret, the `/contexts/{name}/files/...` endpoint returns `has_secret_warning` and `secret_warnings` so operators can move the credential out of the brain pack, but the non-JSON CLI path prints only a success message. In the default warn mode this means `floom contexts push` succeeds while hiding the only warning about the leaked credential unless the user happened to pass `--json`.
```

itachi-hue:
Automated codex review iteration 2

```text
    Surface that loudly at startup instead of letting it fail again."""
    if os.environ.get("COMPOSIO_WEBHOOK_SIGNING_KEY", "").strip():
        return
    count = 0
    try:
        with get_db() as conn:
            row = conn.execute(
                "SELECT COUNT(*) AS cnt FROM workers "
                "WHERE composio_trigger_id IS NOT NULL AND composio_trigger_id != ''"
            ).fetchone()
            count = int(row["cnt"] or 0) if row else 0
    except Exception:
        count = 0
    if count:
        logger.error(
            "COMPOSIO_WEBHOOK_SIGNING_KEY is not configured but %d worker(s) have "
            "enabled Composio event triggers — deliveries are rejected with 503 at "
            "/composio-events and those workers will NEVER fire. Set the key and "
            "register the webhook URL in the Composio dashboard (#908).",
            count,
        )
    elif os.environ.get("COMPOSIO_API_KEY", "").strip():
        logger.warning(
            "COMPOSIO_WEBHOOK_SIGNING_KEY is not configured; Composio event "
            "triggers cannot be enabled until it is set (#908)."
        )


async def require_secret(request: Request) -> str:
    """DEPRECATED: use Depends(get_auth_context) instead."""
    ctx = await get_auth_context(request)
    return ctx.user_id




# ---------------------------------------------------------------------------
# Workspace members (STEP 2 — Codex members design, build-order step 2).
#
# Engine-owned membership endpoints calling the step-1
# ``WorkspaceMemberRepository``. ONE model for BOTH products: the OSS engine is
# the single-owner degenerate case (one active owner = the local user), Cloud
# implements the same Protocol against Supabase. The role matrix lives in the
# repository layer (owner-only role/transfer; owner+admin invite/remove; admin
# cannot target owner/admin); these endpoints map repository PermissionError /
# ValueError to 403 / 400 and never trust the client for authority.
# ---------------------------------------------------------------------------





























def _rate_limit_for_path(path: str) -> tuple[int, float]:
    for pattern, limit in RATE_LIMIT_RULES:
        if pattern.fullmatch(path):
            return limit
    return DEFAULT_RATE_LIMIT


# #1024: PUT /workers/{id}/files — atomic file replace, may carry bundled data.
_WORKER_FILES_PATH_RE = re.compile(r"^/workers/[^/]+/files$")
_WORKER_SOURCE_PATH_RE = re.compile(r"^/workers/[^/]+$")


def _normalized_request_path(request: Request) -> str:
    """Return the route-local path, stripping any ASGI mount prefix.

    B10: a downstream host may mount this engine sub-app under ``/api``
    (``parent.mount("/api", engine.app)``). When mounted, Starlette keeps the
    full prefixed path in ``scope["path"]`` (and ``request.url.path``) while
    recording the mount prefix in ``scope["root_path"]``. The body-size /
    upload exemptions below match route-local paths (``/contexts/...``,
    ``/workers/from-bundle``, ...), so without stripping the prefix every
    contexts/approvals upload in the cloud fell through to the 256 KB JSON
    default and 1 MB images 413'd.

    Stripping ``root_path`` (the canonical ASGI mount mechanism) is
    mount-prefix-agnostic — it works for any mount point and accumulates across
    nested mounts — and is a no-op when not mounted, so the un-prefixed path is
    unchanged. The boundary check (exact match or ``prefix + "/"``) prevents
    over-stripping a sibling segment such as ``/apiary``.
    """
    path = request.scope.get("path") or request.url.path or "/"
    root_path = (request.scope.get("root_path") or "").rstrip("/")
    if root_path and (path == root_path or path.startswith(root_path + "/")):
        return path[len(root_path):] or "/"
    return path


def _body_limit_for_request(request: Request) -> Optional[int]:
    method = request.method.upper()
    if method not in {"POST", "PUT", "PATCH"}:
        return None
    path = _normalized_request_path(request)
    if path == "/workers/from-bundle":
        return FROM_BUNDLE_BODY_LIMIT_BYTES
    if path == "/workspace/import":
        return WORKSPACE_IMPORT_BODY_LIMIT_BYTES
    # Worker source writes may carry data-bundled modules, so they need the same
    # bounded headroom as atomic file deploys without broadening worker run/admin
    # subpaths.
    if (
        (method == "POST" and path == "/workers")
        or (method in {"PUT", "PATCH"} and _WORKER_SOURCE_PATH_RE.match(path))
        or (method == "PUT" and _WORKER_FILES_PATH_RE.match(path))
    ):
        return WORKER_FILES_BODY_LIMIT_BYTES
    if path.startswith("/uploads"):
        return None
    # X4: approval-scoped screenshot uploads stream a multipart image body; the
    # /uploads handler enforces its own size + quota caps, so exempt them from
    # the small JSON default here (authed owner + signed-link public reviewer).
    if path.endswith("/uploads") and (
        path.startswith("/approvals/") or path.startswith("/approvals/public/")
    ):
        return None
    if path.startswith("/contexts"):
        return None
    # #872 FINDING-3: the Slack/WhatsApp webhook handlers advertise (and enforce)
    # a 1 MB cap via channels.common._MAX_WEBHOOK_BODY_BYTES, but the global
    # 256 KB DEFAULT_JSON_BODY_LIMIT_BYTES below fired first — so the advertised
    # 1 MB was a lie and any 256 KB-1 MB Slack event was 413'd before the
    # route's own (signature-gated) size check ran. Exempt them here so each
    # route's explicit 1 MB cap + HMAC verification governs.
    if path in {
        "/slack/events",
        "/slack/commands",
        "/slack/interactivity",
        "/whatsapp/webhook",
    }:
        return None
    return DEFAULT_JSON_BODY_LIMIT_BYTES


def _is_context_upload_request(request: Request) -> bool:
    path = _normalized_request_path(request)
    return (
        request.method.upper() == "POST"
        and path.startswith("/contexts/")
        and path.endswith("/upload")
    )


_WST_ALLOWED_POST_RE = re.compile(r"^/workers/[^/]+/runs$")
_WST_DENIED_PREFIXES = (
    "/secrets", "/connections", "/auth", "/workspace/tokens", "/workspace/secrets",
    "/workspace/settings", "/system", "/settings", "/contexts", "/chat",
)


@app.middleware("http")
async def workspace_token_scope_middleware(request: Request, call_next):
    """Workspace API tokens (wst_) are read+run only.

    Reads stay permission-scoped by the synthetic member actor (shared workers
    only); this gate removes every write surface except firing a run, and
    blocks credential/config surfaces outright — including reads — so a leaked
    token can never enumerate secrets, connections, or settings.
    """
    bearer = (request.headers.get("authorization") or "").strip()
    if bearer.lower().startswith("bearer wst_"):
        path = request.url.path
        if any(path == pfx or path.startswith(pfx + "/") for pfx in _WST_DENIED_PREFIXES):
            return JSONResponse(
                status_code=403,
                content={"detail": "workspace tokens cannot access this resource"},
            )
        if request.method.upper() not in {"GET", "HEAD"} and not (
            request.method.upper() == "POST" and _WST_ALLOWED_POST_RE.match(path)
        ):
            return JSONResponse(

exec
/bin/bash -lc 'git status --short && git ls-files apps/mcp/dist/cli.js | cat' in /home/dev/code_dir/workerOS
 succeeded in 0ms:
exec
/bin/bash -lc 'cat apps/mcp/src/lib/output.ts' in /home/dev/code_dir/workerOS
 succeeded in 0ms:
import chalk from "chalk";

export const log = {
  info: (msg: string): void => { process.stdout.write(`${msg}\n`); },
  ok: (msg: string): void => { process.stdout.write(`${chalk.green("✓ ")}${msg}\n`); },
  warn: (msg: string): void => { process.stderr.write(`${chalk.yellow("! ")}${msg}\n`); },
  err: (msg: string): void => { process.stderr.write(`${chalk.red("✗ ")}${msg}\n`); },
  step: (msg: string): void => { process.stdout.write(`${chalk.dim("· ")}${msg}\n`); },
  heading: (msg: string): void => { process.stdout.write(`\n${chalk.bold(msg)}\n`); },
  kv: (key: string, value: string): void => { process.stdout.write(`  ${chalk.dim(key.padEnd(18))}${value}\n`); },
  blank: (): void => { process.stdout.write("\n"); },
};

export function exitWith(code: number, msg?: string): never {
  if (msg) log.err(msg);
  process.exit(code);
}

type TableColumn<T> = {
  key: keyof T;
  label: string;
};

export function renderTable<T extends Record<string, unknown>>(rows: T[], columns: TableColumn<T>[]): string {
  if (!rows.length) return "(no rows)";
  const widths = columns.map((column) => {
    const maxCell = rows.reduce((max, row) => {
      const value = row[column.key];
      const text = value === undefined || value === null ? "" : String(value);
      return Math.max(max, text.length);
    }, column.label.length);
    return maxCell;
  });
  const header = columns
    .map((column, idx) => column.label.padEnd(widths[idx], " "))
    .join("  ");
  const separator = columns.map((_, idx) => "-".repeat(widths[idx])).join("  ");
  const body = rows.map((row) =>
    columns
      .map((column, idx) => {
        const value = row[column.key];
        const text = value === undefined || value === null ? "" : String(value);
        return text.padEnd(widths[idx], " ");
      })
      .join("  "));
  return [header, separator, ...body].join("\n");
}

export function printJson(value: unknown): void {
  console.log(JSON.stringify(value, null, 2));
}

exec
/bin/bash -lc "cd apps/mcp && node -e \"import('./dist/cli.js').then(()=>{})\"" in /home/dev/code_dir/workerOS
 succeeded in 497ms:
codex
The new contexts CLI commands align with the existing /contexts API surface, build and tests pass, and I did not find a discrete regression introduced by the patch.
The new contexts CLI commands align with the existing /contexts API surface, build and tests pass, and I did not find a discrete regression introduced by the patch.
```

itachi-hue:
Closing after triage. This PR adds MCP context CLI commands, but it does not resolve the linked P0 cloud-onboarding epic (#1797). The underlying issue should remain open and be handled by a correctly scoped follow-up instead of merging this as a fix for #1797.

Issue #1078, opened 2026-06-14
[cloud #272] Context file version-history never recorded + WhatsApp binding store SQLite-backed on cloud


itachi-hue:
Originating cloud issue: floomhq/workeros-cloud#272 (item 2, PUT /workers/{id}/files partial edit, was fixed cloud-side). Remaining engine items:

1. **Context version-history never recorded.** POST /contexts/X then two PUT /contexts/X/files/notes.md (different content) -> GET .../versions returns []. The knowledge-pack versioning/rollback feature records no history despite multiple writes. (Minor: a freshly created context returns "writeable": false in the create response.)
3. **WhatsApp binding store is SQLite-backed.** engine/apps/api/channels/whatsapp.py uses `from db import get_db` for sender bindings, which won't persist across redeploys / scale across instances on the ephemeral cloud. The binding-store seam (#1007) should be the path; ensure cloud can register a Supabase-backed binding store (same class as #233/#269).

itachi-hue:
Investigated both engine items.

**Item 1 (context version-history) — working as designed; not a recording bug.**
I added an empirical test (`apps/api/tests/test_1078_context_versions_probe.py`) that creates a context, writes `notes.md` twice, and reads `/versions`:
- **Non-sensitive** writeable context → **2 versions recorded** (git history works: `get_log --follow` over the committed pack file).
- **Sensitive** context (the create default, `ContextCreateRequest.sensitive=True`) → **[]**, because `_git_commit_context` early-returns for sensitive packs (`is_context_sensitive` — they may hold credentials and must never enter git).

So the cloud #272 repro returned `[]` because the pack was **sensitive-by-default** (and on the ephemeral cloud the git workspace is also non-persistent / a Supabase-backed store, so there is no git log to read). Recording history for sensitive packs would break the deliberate secret-safety guarantee. Whether a *writeable* pack should default to `sensitive=false` (and thus be versioned) is a **product decision**, not a code bug — leaving that for a product call. The minor `writeable:false` on the create response is the same default-flag behavior.

**Item 3 (WhatsApp binding store is SQLite-backed) — cloud seam, same class as #1075/#1007.**
This is a new pluggable-store seam so cloud can register a Supabase-backed binding store (no SQLite-engine bug to fix/test here). Tracking it with the seam work in #1075 / #1007.

Keeping this **open** — no OSS-testable engine defect remains; the actionable parts are a product decision (item 1) and a cloud seam (item 3). cc @vivek

[...]

PR #1867, opened 2026-06-22
Fix CLI cloud smoke regressions

### LINKED ISSUE #1863: [P1][cloud][cli-auth] Approval page can show approved from wrong browser/account even though CLI is not approved
﻿## Summary

The cloud CLI device-approval page can display an "approved" success state when the approval link is opened from a different browser/account, but the CLI login is not actually approved for the initiating session/account.

This is misleading and makes users think login succeeded when the CLI remains unauthenticated or stuck polling.

## Repro

1. Start a cloud CLI login from one environment:
   `npx @floomhq/floom@5.0.2 login --cloud`
2. Copy/open the approval link in a different browser profile or a browser logged into a different Floom account/session.
3. Complete whatever UI action is shown.

Actual:
- The frontend shows an approved/success state.
- The originating CLI login is not actually approved for the intended account/session.

Expected:
- If the browser session is not authenticated, redirect to login with the device code preserved.
- If the browser session is authenticated as the wrong account or cannot approve that device code, show a clear error like:
  - "Sign in to the account that started this CLI login"
  - "This approval link is not valid for the current account"
  - "Approval failed; return to the CLI and start a new login"
- Do not show an "approved" state unless the backend confirms the device code was approved and bound to the correct account.

## Impact

This directly affects cloud onboarding. It looks like login succeeded but leaves the CLI unapproved, which is especially confusing when users have multiple browser profiles/accounts.

## Notes

Observed live on 2026-06-22 during CLI onboarding smoke. The user reported: opening the approval link from a different account/browser still says approved, but is not actually approved.

This likely needs the frontend to treat non-OK / wrong-account / unauthenticated responses as explicit failure states instead of rendering success optimistically.

### LINKED ISSUE #1864: [P2][cli] run command prints Node libuv assertion after expected API 409
﻿## Summary

When `floom run` receives an expected API error for a paused worker, it prints the API error and then a Node/libuv assertion failure.

## Repro

CLI: `npx @floomhq/floom@5.0.2`
Workspace: live cloud `vbellala`
Worker: paused `codex-cli-smoke-vbellala`

Command:
`floom run codex-cli-smoke-vbellala --input text=cli-smoke --json`

Output:
```
floom failed: API POST /workers/codex-cli-smoke-vbellala/runs failed with HTTP 409: This worker is paused. Turn it on to run it again.
Assertion failed: !(handle->flags & UV_HANDLE_CLOSING), file src\win\async.c, line 76
```

Expected:
- CLI exits non-zero with the clean API error only.
- No Node assertion/native runtime failure should be printed.

## Impact

This makes normal API errors look like CLI/runtime crashes and is especially confusing during onboarding smoke tests.

### LINKED ISSUE #1865: [P1][cloud][workers] workers enable reports success but worker remains paused
﻿## Summary

On live cloud, `floom workers enable <id>` reports success, but the worker remains `enabled: false` and manual runs are still rejected as paused.

## Repro

CLI: `npx @floomhq/floom@5.0.2`
Workspace: `vbellala` (`ws_7ff0773b15684f`)
Worker: throwaway `codex-cli-smoke-vbellala`

1. Create/push a manual worker.
2. Run `floom workers disable codex-cli-smoke-vbellala --json`.
   - Response shows `enabled: false`.
3. Run `floom workers enable codex-cli-smoke-vbellala`.
   - CLI prints `✓ Enabled codex-cli-smoke-vbellala`.
4. Run `floom workers show codex-cli-smoke-vbellala --json`.

Actual:
- `enabled` is still `false`.
- `status` remains `needs_attention`.
- `floom run codex-cli-smoke-vbellala --input text=cli-smoke --json` fails with HTTP 409: `This worker is paused. Turn it on to run it again.`

Expected:
- `workers enable` should persist `enabled: true` and manual runs should no longer return paused/409.
- If enable cannot apply, CLI/API should return a non-zero error instead of success.

## Impact

A user can pause a worker from the CLI but cannot resume it from the CLI, even though the CLI reports success.

### LINKED ISSUE #1866: [P2][cli][windows] completion powershell is unsupported on the default Windows shell
## Summary

On Windows, the Floom CLI runs under PowerShell by default, but loom completion powershell is rejected. The command only accepts ash, zsh, and ish.

## Repro

`powershell
npx --yes @floomhq/floom@5.0.2 completion powershell
`

## Actual

`	ext
floom failed: Shell must be one of: bash, zsh, fish
`

## Expected

Either generate PowerShell completion, or document that shell completion is Unix-shell only and avoid presenting completion <shell> as a generally available CLI path on Windows.

## Notes

completion bash does work and prints a completion script. This is not a cloud runtime blocker, but it is a visible CLI path gap for Windows users.

itachi-hue:
﻿## Summary

Fixes the concrete regressions found during the live CLI cloud smoke:

- Fixes #1865 by making pause/resume updates write through the worker row owner, so authorized workspace-admin mutations do not pass authz but no-op under the acting admin id.
- Fixes #1864 by handling expected `floom run` API rejections such as paused-worker HTTP 409 as clean CLI errors instead of throwing to the top-level crash handler.
- Fixes #1863 by requiring the CLI-auth approval page to receive JSON `{ ok: true }` before showing the terminal Approved state; non-JSON/proxy/login-shell responses stay on the action page with an account/login error.
- Fixes #1866 by adding `completion powershell` / `completion pwsh` support.
- Improves #1078 UX by explaining that empty context version history is expected for sensitive brain packs and pointing users at `contexts create --no-sensitive` for rollback/versioned packs.
- Closes the review follow-up by scoping the admin mutation fallback to active workspace membership before using the global `get_any` worker row.

## Tests

- `python -m pytest apps/api/tests/test_1463_worker_mutation_authz.py apps/api/tests/test_worker_pause_resume.py -q`
- `cd apps/mcp && npm test -- --test-name-pattern "completion|workers run reports paused|contexts versions"`
- `cd apps/web && npm ci && npm test -- --run tests/cli-auth-seams.dom.test.tsx`
- `git diff --check`

## Notes

The admin mutation follow-up adds direct tests for same-workspace admin allowed and cross-workspace admin denied. A broader mixed API run hit pre-existing fixture startup/auth failures unrelated to this helper (`FLOOM_SECRET` / admin-only setup), so the targeted mutation/lifecycle suite is the authoritative validation for this follow-up.

`apps/mcp` targeted test command ended up exercising the full MCP test suite because of the node test pattern matching behavior; all 116 tests passed.


itachi-hue:
Fixed the review catch in commit 601e44f2.

The admin mutation workspace-membership check now calls the real `WorkspaceMemberRepository.get(workspace_id=..., user_id=...)` method and requires `status == "active"`; it no longer calls the non-existent `get_role` method. Tests were updated so the fake mirrors the real `get(...)` contract, plus a SQLite-backed regression was added through the real member repository.

Verification:
- `python -m pytest apps/api/tests/test_1463_worker_mutation_authz.py apps/api/tests/db/test_workspace_members_and_visibility.py -q` passed: 23 tests
- `rg -n "get_role" apps/api tests` returned no matches
- `git diff --check` passed

PR #1922, opened 2026-06-23
Fix launch-blocking OSS regressions

### LINKED ISSUE #1913: CLI workers run --json emits progress text before JSON
## Summary

`floom workers run <id> --json` prints progress/status lines before the JSON payload, so stdout is not valid JSON even though the command is explicitly in JSON mode.

## Live prod evidence

Environment:
- API: `https://workeros-api.floom.dev`
- CLI: global `floom --version` = `5.0.2`
- Date: `2026-06-23`
- Workspace: `ws_7ff0773b15684f`

During a live destructive CLI smoke, I created `codex-cli-live-520873`, then ran:

```bash
floom workers run codex-cli-live-520873 --input text=smoke --json
```

The worker run itself succeeded and produced the expected output, but stdout began with progress text before the JSON object:

```text
· Run started: run_e21efafd7739
· Status: running
· Status: completed
{ "id": "run_e21efafd7739", "worker_id": "codex-cli-live-520873", ... }
```

A JSON parser fails immediately:

```text
Unexpected token '·', "· Run star"... is not valid JSON
```

## Expected

With `--json`, stdout should contain exactly one parseable JSON document. Progress/status logs should be suppressed or sent to stderr.

## Actual

Human progress lines are emitted to stdout before the JSON payload.

## Impact

Automation cannot reliably consume `workers run --json`; it has to strip human output first. This is especially painful for CI or agent integrations that need the run id/output from stdout.

## Cleanup

The test worker was deleted successfully after the run.

### LINKED ISSUE #1916: [HIGH] Declared+set worker secret not injected into runs in a newly-created workspace (missing_secrets:[] but run sees it missing)
## Problem
A worker (`staging-gate`) in a freshly-created workspace (`floom product`, ws_90adf1e0eba542) declares `STAGING_GATE_SSH_KEY` + `STAGING_GATE_CALLBACK_SECRET` (in both `capabilities.secrets` and `exec.secrets`). Both are SET (`workeros secrets set` returned ✓; `workers show --json` reports `missing_secrets: []`). Yet every run fails with the worker seeing the secret as ABSENT — `secrets.json`/env does not contain it.

## Evidence
- `secrets list` shows both names; `missing_secrets: []`.
- run_cb640197f375 and run_fb93af48bd87: `{"status":"error","summary":"Missing STAGING_GATE_SSH_KEY secret; no audit or merge ran."}` — run.py read `secrets.json`/env and the key was not present.
- Contrast: the same native-secret pattern WORKS for another worker (`content-publish`) in a different, older workspace (Nova Search) — its declared BUFFER_API_TOKEN injects fine. So this looks **specific to the newly-created workspace** (or to multi-secret declaration), i.e. a regression/edge of the #1730/#1763 secret-injection fix.

## Impact
Any worker in a new workspace that needs a declared secret can't get it at runtime, even though the UI/API says it's set. Blocks e.g. a worker that must SSH/call out with a credential.

## Ask
Ensure declared+set secrets are actually materialized into the run (secrets.json/env) for newly-created workspaces, not just reported as present. Possibly the workspace secret store isn't fully wired to the run-secret materialization path for new workspaces.

### LINKED ISSUE #1920: [HIGH] New worker runs execute a STALE cached build, not the latest pushed source
## Problem
After `workeros workers push`, the worker's stored source IS updated (server returns the new run.py), but **new runs execute an OLDER cached version** of the code.

## Evidence (workspace 'floom product' ws_90adf1e0eba542, worker staging-gate)
1. Pushed run.py with a new `ssh_key_b64` input fallback. `GET /workers/staging-gate` confirms the server source contains the fallback (`ssh_key_b64` present, len 11627).
2. Triggered a run passing `ssh_key_b64=<base64 key>`. The run record shows `input_values.ssh_key_b64` present (non-empty).
3. The run's output was the OLD code path: `{"error":"missing STAGING_GATE_SSH_KEY"}` — i.e. it never executed the new `inputs.get("ssh_key_b64")` fallback branch that exists in the current server source.

So: stored source = new, executed code = old. New runs do not pick up the latest pushed worker code.

## Likely mechanism
Per-worker e2b sandbox image/template (or worker version) is built/cached and a fresh `push` doesn't rebuild/invalidate it, so runs keep using the prior build. (Same class would affect any iterative worker dev.)

## Impact
Iterating on a worker (edit -> push -> run) silently runs stale code, making worker development/debugging very confusing — changes appear to have no effect. Discovered while iterating on a staging-gate worker.

## Ask
Ensure `workers push` invalidates/rebuilds the execution image so the next run uses the latest source, or expose explicit versioning + a way to run the latest. At minimum document the cache + how to force a rebuild.

### LINKED ISSUE #1921: bug: worker-author auto-register fails with trigger schema list-vs-dict mismatch
## Repro

In `secretary` workspace under `fede@floom.dev`, asked Emily to create a scheduled worker. After 3 attempts the worker-author agent produces bundles that consistently fail auto-register:

```
[warning] Could not auto-register the drafted worker:
400: {'message': 'Schema validation failed', 'errors': [{'loc': 'trigger', 'msg': 'Input should be a valid dictionary or instance of WorkerContractTrigger', 'type': 'model_type'}]}
```

## Root cause

Generated bundle YAML has trigger as a LIST:

```yaml
trigger:
- type: schedule
  cron: 30 5 * * *
```

But the server-side `WorkerContractTrigger` schema expects a single DICT:

```yaml
trigger:
  type: schedule
  cron: 30 5 * * *
```

(no leading dash)

Either:
- (a) the worker-author LLM prompt needs the schema example fixed to be a dict not a list
- (b) the server validator should accept both forms (single-element list → dict coerce)

## Runs

- run_83d9d08e8c91 — attempt 1, drafted bundle with cron `30 6 * * *` (wrong DST), didn't try register
- run_422a5686ef31 — attempt 2, fixed cron to `30 5 * * *`, auto-register FAILED with the schema error above
- run_1ed4194d1608 — attempt 3 with explicit "fix as dict not list" instruction, ALSO not landing in workers list (status: 4 workers, no ec261-flight-claim-watch)

## Impact

Users (incl. Federico) cannot complete the "Emily creates a scheduled worker" path without manual server bundle edit. Pure worker-author flow is broken for schedule triggers.

## Quick fix suggestion

Worker-author prompt template — change trigger example from list to dict.

itachi-hue:
## Summary

Fixes the launch-blocking OSS regressions:

- #1920: prevent stale E2B warm-pool reuse when DB `runtime.bundle_sha256` is stale relative to the pushed worker directory
- #1916: check required secrets against the worker owner scope, matching the sandbox injection scope
- #1921: accept and repair worker-author single-item `trigger:` lists into the schema's mapping shape, and tell worker-author not to emit list-shaped triggers
- #1913: keep `workers run --json` stdout as a single parseable JSON document by suppressing progress lines in JSON mode

Closes #1920
Closes #1916
Closes #1921
Closes #1913

## Tests

- `python -m pytest apps/api/tests/test_trigger_type_aliases.py apps/api/tests/test_1137_worker_author_provider.py apps/api/tests/test_batch7_backend.py apps/api/tests/test_1433_context_mount_selection.py -q`
- `python -m pytest apps/api/tests/db/test_scheduled_run_secret_scope.py apps/api/tests/db/test_1730_capabilities_secrets_injected.py apps/api/tests/test_e2b_declared_secret_env_concurrency.py -q`
- `python -m pytest apps/api/tests/test_run_perf_timing.py apps/api/tests/test_1735_bundle_sha_canonicalization.py -q`
- `cd apps/mcp && npm test -- --runInBand test/cli-workers.test.js`


itachi-hue:
Follow-up on the perf verify: the first version compared the trusted DB bundle sha to the full _hash_tree digest, but DB _files uses a length-framed text-file hash. Updated the trust check to hash the same embedded source-file set/algorithm as _bundle_sha256_from_manifest_files, and only fall back to full disk tree hashing when that source hash mismatches or is unavailable.\n\nAdditional tests passed after the amend:\n- python -m pytest apps/api/tests/test_1433_context_mount_selection.py apps/api/tests/test_run_perf_timing.py apps/api/tests/test_1735_bundle_sha_canonicalization.py -q -> 33 passed\n- python -m pytest apps/api/tests/test_trigger_type_aliases.py apps/api/tests/test_1137_worker_author_provider.py apps/api/tests/test_batch7_backend.py -q -> 25 passed

Issue #916, opened 2026-06-12
[MEDIUM] Worker-call run tokens do not check user status


federicodeponte:
## Problem

`MultiMemberAuthProvider._verify_worker_call_token()` decodes and validates the HMAC-signed worker-call run token but never checks whether the encoded `user_id` is still active. A disabled or deleted user whose parent run issued child-run tokens can keep invoking callable workers until the token expires.

## Code

`apps/api/auth/multi_member.py:153-167`:

```python
async def _verify_worker_call_token(self, raw_token: str) -> AuthContext:
    payload = validate_worker_call_token(raw_token)
    user_id = str(payload.get("user_id") or "")
    if not user_id:
        raise HTTPException(status_code=401, detail="invalid run token: missing user_id")
    return AuthContext(user_id=user_id, role="member", auth_method="run_token", ...)
```

## Impact

A disabled/deleted user's in-flight parent run can continue to issue child-run tokens and trigger downstream workers for up to the token lifetime (default 1h).

## Fix

Validate `user_id` against `users.disabled` / existence before returning the `AuthContext`. Treat missing or disabled users the same as an invalid token.

## Severity

Medium (short-lived token, but user-lifecycle bypass)
Area: backend, security, auth
Reporter: security audit

itachi-hue:
**Fixed in PR https://github.com/floomhq/workeros/pull/962 (branch `g1`).**

**RCA.** `_verify_worker_call_token()` trusted the HMAC payload's `user_id` without checking whether that user still exists or is enabled — a disabled user's in-flight parent run could keep minting child-run access for the token lifetime (default 1h).

**Fix.** The same `_require_active_token_user()` lifecycle check from #915 now runs in two places: inside the auth provider, and at the **middleware perimeter** (`auth_middleware` validates `wrt_` tokens before routing, so the check fires even on the containment fast-path).

**Test** (`tests/test_g1_auth_lifecycle.py::test_run_token_rejected_when_user_disabled`): active user's run token → 403 from the normal containment rule; after `PATCH /users/{id} {"disabled": true}` the same token → 401. Green.

PR #1925, opened 2026-06-23
fix(api): accept script worker-call run tokens

federicodeponte:
## Summary

Fixes the script-mode Composio proxy token mismatch behind cloud bug #608.

Script workers that declare `calls:` receive a `wrt_` worker-call token in `WORKEROS_RUN_TOKEN`. The worker sends that value through `X-Floom-Run-Token` to `/runs/{run_id}/composio-execute/{tool}`, but the middleware and proxy handler only accepted the older simple `run:` token family. That rejected the request with 401 before the proxy executed.

This change resolves the run id from either token family for run-scoped platform proxy callbacks:

- simple `run:` token -> `run_id`
- `wrt_` worker-call token -> `parent_run_id`

The existing path allowlist and run-id equality checks remain in place.

## Tests

- `cd engine/apps/api && python3 -m pytest tests/test_composio_proxy.py -q`
- `cd engine/apps/api && python3 -m pytest tests/test_run_token.py tests/test_worker_calls.py -q`


itachi-hue (approved):
Reviewed and pushed a small hardening commit (2aeacd55) on top of this PR. The original fix correctly routes script-mode worker-call run tokens through the run-scoped platform proxy path; the extra commit preserves the existing #916 lifecycle invariant by rejecting wrt_ run tokens whose owning user has been disabled or deleted.\n\nValidated locally:\n- python -m pytest apps/api/tests/test_composio_proxy.py -q\n- python -m pytest apps/api/tests/test_composio_proxy.py::test_composio_proxy_accepts_script_worker_call_token_header apps/api/tests/test_managed_llm_proxy.py::test_worker_call_token_can_only_use_parent_run_managed_llm -q\n\nCI note: the previous runtime failures were in tests/test_418_approval_gate.py::Test418ApprovalGate::test_phase1_withholds_secrets_until_approved_follow_up on both Linux and Windows, outside this diff.

PR #2025, opened 2026-06-26
fix(api): reject credential run inputs

itachi-hue:
## Summary
- reject worker manifests that expose API keys, tokens, passwords, private keys, or auth-flavored connection fields as run inputs
- clarify Emily worker-authoring/tool rules to use top-level connections and declared secrets instead of credential inputs
- add regression coverage that still allows business fields like Linear team_key

## Tests
- python -m pytest apps/api/tests/test_worker_credential_input_guard.py apps/api/tests/test_673_674_675_emily_worker_authoring.py apps/api/tests/test_683_emily_agent_mode_routing.py -q


itachi-hue:
Pushed a follow-up commit (7ffdd5b6) to fix a false-positive: the strong `_CREDENTIAL_INPUT_RE` was matched against name+label+placeholder+description concatenated, so a legitimate business input whose help text merely *mentions* credentials/password/api key (e.g. a `summary` described "include candidate credentials and skills") was wrongly rejected and couldn't be saved. Now the strong pattern matches the field **identity** (name+label) only; placeholder/description feed just the weaker auth-hint heuristic (the same split the connection branch already uses). Added a regression test. Verified the existing reject-tests still pass (they're name/label-driven).

PR #2032, opened 2026-06-26
fix(web): collection loading and settings IA cleanup

itachi-hue:
## Summary
- make collection route loading server-safe so Workers/Runs/Connections/Approvals/Library do not hit the shared 'Something went wrong' route error while loading
- stream Library initial data like Workers/Runs/Connections instead of blocking the server render on /contexts
- restore the historical 8-item Settings IA: six workspace rows and two account rows
- consolidate personal tokens, workspace token, API, MCP, CLI, and Git under Account > Developer
- tighten collection search toolbar spacing and remove the fake secret reveal control

## Tests
- npm test -- --run tests/settings-nav-groups.test.ts tests/fl-batch-14.test.ts
- npm test -- --run tests/settings-nav-groups.test.ts tests/fl-batch-14.test.ts tests/collection-loading-server-safe.test.ts tests/collection-loading-static-header.dom.test.tsx tests/collection-skeleton.dom.test.tsx tests/library-streamed-initial-data.test.ts tests/collection-toolbar-layout.test.ts tests/secret-value-field.test.ts
- npx eslint app/settings/page.tsx lib/settings/nav-groups.ts tests/settings-nav-groups.test.ts tests/fl-batch-14.test.ts


itachi-hue:
﻿One regression in the settings consolidation: the new Developer inner tabs are not tied to URL/sub-tab state, so old links and the new cross-links route to the Developer section but always show the default Personal tokens tab.

Specifically, the new `Tabs` in `DeveloperSection` is uncontrolled and always defaults to `personal-tokens`, but the migration now maps every old programmatic-access route (`?tab=api`, `?tab=git`, `?sel=connect`, `?sel=workspace_token`, `?sel=personal_tokens`, etc.) to only the outer `developer` section. Because `subTabKeysForSection()` still returns sub-tabs only for `system`, none of those aliases can select the intended inner Developer tab, and the new in-pane cross-links also call `navigateSettingsSelection("developer")`, which is a no-op when already in Developer and leaves users on the current/default tab. For example, a prior API/Git deep link now opens Personal tokens, and the "Workspace token" cross-link from Personal tokens does not switch tabs. The Developer tab state needs to be represented in `CollectionState.tab`/`?tab=` (or otherwise controlled), with legacy aliases translated to the matching inner tab.

Validation run locally in an isolated PR worktree:
`npm test -- --run tests/settings-nav-groups.test.ts tests/fl-batch-14.test.ts tests/collection-loading-server-safe.test.ts tests/collection-loading-static-header.dom.test.tsx tests/collection-skeleton.dom.test.tsx tests/library-streamed-initial-data.test.ts tests/collection-toolbar-layout.test.ts tests/secret-value-field.test.ts` passed: 8 files, 26 tests.

itachi-hue:
Closing — superseded. The collection-loading / toolbar / secret-reveal pieces of this PR had already landed (#2030/#2028/#2022), and the BrainCollection change would have regressed #2029. The one net-new piece — the 8-item Settings IA — was salvaged cleanly onto current main and merged as #2035.

PR #2054, opened 2026-06-27
Revert agent copy back to workers

itachi-hue:
## Summary
- revert PR #2039 (`v1-workers-to-agents`) so product copy returns from "agents" back to "workers"
- restores the matching web test/docs copy updates from that PR

## Tests
- `git diff --check HEAD~1..HEAD`
- Attempted `npx vitest run tests/workspace-switcher-copy.test.ts tests/medium-issue-batch-source.test.ts tests/collection-pages.dom.test.tsx tests/runtime-labels.test.ts` from `apps/web/`, but the checkout had no installed web deps (`vitest/config`, `@vitejs/plugin-react`).
- Attempted `npm ci` in `apps/web/` to install deps; failed with `ENOSPC: no space left on device` before install completed. Removed the partial `node_modules` afterward; branch is clean.


itachi-hue:
Added a README follow-up to define the terminology clearly:

> Floom turns recurring work into workers that run on a schedule, webhook, or button click with approvals, logs, outputs, and every run on the record.
> A worker is the complete job package: readable YAML (`worker.yml`) plus optional `run.py`, `SKILL.md`, attached memory, scoped secrets, and approved tools.

Rationale: we should not use "worker" as a softer synonym for "agent". In Floom, a worker is the whole deployable/auditable unit: manifest, runtime entrypoint, memory/context, secrets, tools, trigger, approvals, and run history. Some workers use an agent loop, but the worker is bigger than the agent.

PR #2114, opened 2026-07-01
fix(web): add first-worker quick start empty state

itachi-hue:
## Summary
- Replace the workers empty-state copy with a "Create your first worker" onboarding panel.
- Add quick-start guidance for using a coding agent with the Floom CLI or MCP.
- Include concrete example prompts and install/login/MCP commands.
- Add a source-level regression test for the empty-state copy and commands.

## Verification
- `git diff --check`
- `node` source check for the required empty-state copy and stale-copy removal

## Not run
- `npx vitest run tests/workers-empty-state-quick-start.test.ts` could not start locally because this checkout is missing web node dependencies (`vitest/config`, `@vitejs/plugin-react`).
- `npm run lint -- --file app/workers/WorkersCollection.tsx --file tests/workers-empty-state-quick-start.test.ts` could not start locally because `eslint` is not installed in this checkout.


itachi-hue:
Updated after review: CollectionView now distinguishes true empty collections from filtered-empty results via states.filteredEmpty. The workers page only shows the first-worker CLI/MCP quick start when items.length === 0; search/filter zero-result states now show "No workers found" and keep the toolbar. Added DOM coverage for the filtered-empty case plus the workers quick-start source check.

Checks run locally:
- git diff --check
- Node source assertion for workers quick-start/filtered-empty copy

Vitest/ESLint still cannot run in this checkout because apps/web/node_modules is missing the test/lint tooling.

itachi-hue:
Superseded by #2124, which folds in this first-worker empty-state work along with the broader actionable feedback changes.

Issue #2269, opened 2026-07-17
workspace_chat conversation contamination: no conversation_id bleeds into unrelated conversation; passing a real conversation_id forks a new one instead of resuming


federicodeponte:
## Summary
Two related conversation-continuity bugs in `workspace_chat`:
1. A message sent with **no** `conversation_id` was appended into an **unrelated, pre-existing conversation** instead of starting a fresh one.
2. Passing a **real, valid** `conversation_id` to resume a conversation instead created a **brand-new conversation** — the model responded "no prior question to reference," even though the target conversation had 20 prior messages.

Confirmed via `conversations_list` message-count diffing before/after each call: the target conversation's message count stayed frozen in case 2 (proving the reply did NOT land there), while in case 1 an unrelated conversation's count changed unexpectedly.

## Repro
1. Note message count of an existing, unrelated conversation via `conversations_list`.
2. Call `workspace_chat` with a new message and no `conversation_id`.
3. Re-check `conversations_list` → the unrelated conversation's message count changed (message landed there instead of a new conversation).
4. Separately: pick a real conversation with 20 messages, call `workspace_chat` passing its `conversation_id`.
5. Response: "no prior question to reference" — indicates a fresh conversation was created despite passing a valid id, and the target conversation's message count in `conversations_list` stayed unchanged (no append happened).

## Expected
- No `conversation_id` → starts a new, isolated conversation.
- Valid `conversation_id` → resumes that exact conversation with full prior context.

## Actual
Both directions are broken/inverted: omitting the id causes cross-conversation bleed; providing it causes a fork instead of a resume.

## Severity
P0 — cross-session data bleed risk (one user/session's message can land in another conversation's history) plus complete failure of multi-turn conversation continuity.

## Security note
Filed with `area:security` because unrelated-conversation-append is a session isolation / data bleed concern, not just a UX bug — needs check for whether this crosses workspace or user boundaries, not just conversation boundaries within one workspace.

_Filed from 2026-07-17 WorkerOS launch-readiness audit._

federicodeponte:
Fixed and merged via #2287 conversation isolation fix. Closing (squash-merge title referenced but did not auto-close).

Issue #2266, opened 2026-07-17
workspace_chat returns -32603 Internal error while the run succeeds server-side (false failure to caller)


federicodeponte:
## Summary
`workspace_chat` returned a JSON-RPC `-32603 Internal error` to the MCP client, but the underlying run actually completed successfully server-side (status `completed`, valid output produced, $0.053 billed). A naive agent/caller has no way to know the call actually worked and will report false failure (and may retry, causing duplicate billed runs). A second call also intermittently surfaced a distinct `agent_runtime_error`.

## Repro
1. Call `workspace_chat` with a normal prompt.
2. Client receives `-32603 Internal error`.
3. Cross-check via `runs_get`/billing: the run is `completed`, output is present and valid, $0.053 was billed.
4. Repeat call: intermittent `agent_runtime_error` surfaces on a second attempt.

## Expected
`workspace_chat` returns the successful result (or at minimum a status that matches the true run state) when the run completed.

## Actual
Client-visible error despite a successful, billed run underneath — worst case, this causes callers to double-run (double-bill) on retry.

## Severity
P0 — false-failure signal on the primary chat entrypoint, with a billing/duplication risk on retry.

## Additional evidence
Intermittent `agent_runtime_error` on repeat call — may be a related/same root cause (unstable response path from the underlying chat/run mediation) rather than a separate bug; worth investigating together.

_Filed from 2026-07-17 WorkerOS launch-readiness audit._

federicodeponte:
Fixed and merged via #2287 workspace_chat false-failure fix. Closing (squash-merge title referenced but did not auto-close).

PR #2287, opened 2026-07-19
fix(mcp): workspace_chat false-failure + conversation isolation (#2266, #2269)

federicodeponte:
## Root causes

**#2266 — workspace_chat returned MCP -32603 while the run actually succeeded**
- `apps/api/chat_service.py` `stream_chat()`: after the reply was fully generated and the `finish` event was about to be emitted, several post-completion side effects (`_persist_tool_card`, `insert_message` for tool/assistant messages, `_sanitize_preview_text`/`_ensure_bare_greeting_identity`, `_maybe_evict_conversation`, MCP server teardown, provider client `.aclose()`) ran un-guarded. Any exception there propagated out of the coroutine and was surfaced by the MCP layer as a bare `-32603 Internal error`, discarding the already-successful (and already-billed) reply.
- `apps/api/channels/common.py` `collect_agent_reply(...)`: after the `finish` part was consumed off the queue, `await task` could still raise (client teardown, provider close) — that late exception converted a completed run into a failure for the caller.

**#2269 — conversation isolation broken**
- `apps/api/main.py` `_workspace_agent_mcp_conversation_id()`: with no `conversation_id` supplied, the old code fell back to a fixed `"default"` value, so every no-id MCP call from every session mapped to the *same* `"langdock:default"` thread for a given user — cross-session bleed into an unrelated conversation.
- Same function: when a real internal id (`conv_...`, e.g. one returned from a previous reply or `conversations.list`) was passed in to resume a thread, it was run through a prefix+hash transform (`langdock:<hash>`) instead of being used directly, which silently forked a brand-new, empty conversation instead of resuming.

## Fixes

- `apps/api/chat_service.py`: wrapped every post-run side effect (tool card/transcript persistence, reply sanitization, assistant message persistence, conversation eviction, MCP teardown, provider close) in `_persist_side_effect` / local `try/except`, logging failures instead of propagating them. The `finish` event is emitted with the real reply regardless of downstream bookkeeping failures.
- `apps/api/channels/common.py`: added `AgentReplyError` (client-safe pre-completion failure) and `collect_agent_reply_details(...)`, which returns `{reply, conversation_id, message_id}` from the real `finish` event; a late exception from `await task` *after* `finish` was already observed is logged and ignored rather than re-raised.
- `apps/api/main.py`:
  - `_workspace_agent_mcp_conversation_id()`: no/empty id → `None` (stream_chat starts a fresh conversation); `conv_...` ids pass through untouched so the caller resumes the exact conversation; other stable client ids keep the legacy deterministic `langdock:<safe>` mapping for backwards compatibility.
  - New `_mcp_workspace_chat_result()` shared handler (used by both MCP dispatch paths): validates a supplied `conv_...` id up front via `resolve_conversation_id`/`get_conversation` and returns a clear tool error (`isError: true`) if it doesn't exist for/belong to the caller's workspace, instead of silently forking a new conversation. Genuine chat failures are caught and returned as an actionable `isError` tool result (client-safe message via `safe_llm_error_message`) instead of a bare JSON-RPC `-32603`.
  - `workspace.chat` tool schema/description updated to document the omit-vs-resume semantics.

## Tests

New `tests/test_workspace_chat_mcp_2266_2269.py` (mirrored under `apps/api/tests/`) covers:
- `test_workspace_chat_no_id_creates_distinct_conversations` — two no-id calls produce two distinct conversations (no cross-session bleed).
- `test_workspace_chat_resumes_existing_conversation_by_id` — a call with a real `conv_...` id appends to and sees the history of that conversation.
- `test_workspace_chat_unknown_conversation_id_errors_without_forking` — an id that doesn't exist/isn't owned errors clearly instead of forking.
- `test_workspace_chat_returns_reply_when_post_run_bookkeeping_fails` / `test_workspace_chat_returns_reply_when_post_finish_close_fails` — a run that completes still returns its reply when post-run persistence/teardown throws (no `-32603`).
- `test_workspace_chat_genuine_failure_surfaces_actionable_message` — a real pre-completion failure still surfaces as an actionable error.
- `test_mcp_conversation_id_mapping_unit`, `test_workspace_chat_serve_endpoint_roundtrip` — unit/roundtrip coverage of the id-mapping function and the MCP endpoint.

Existing `tests/test_langdock_workspace_agent_mcp.py` updated for the new return contract and re-verified.

Targeted run (`python -m pytest tests/test_workspace_chat_mcp_2266_2269.py tests/test_langdock_workspace_agent_mcp.py -v`): **14 passed**.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


federicodeponte:
## Follow-up fix (commit ee00d2310)

An independent codex review of the initial diff found a real gap: `collect_agent_reply_details` (apps/api/channels/common.py) only awaited `queue.get()` in its receive loop. `stream_chat` performs conversation resolution/creation/history-load/provider setup **before** it enters the try block that emits `error`/`finish` parts. An exception in that pre-event section left the stream task dead without ever putting anything on the queue, so the collector hung forever instead of surfacing an actionable error — a regression risk directly in scope of #2266 ("genuine pre-completion errors must surface as an actionable message").

Fix: the receive loop now races the next queued part against stream-task completion (`asyncio.wait(..., return_when=FIRST_COMPLETED)`). If the task finishes before a `finish`/`error` part is observed, any last-moment queued part is drained first, then the task's exception is propagated (or `AgentReplyError` is raised if the task returned cleanly without ever emitting `finish`). Post-finish tolerance is unchanged — once `finish` has been observed, a later task exception is still only logged.

Added regression coverage (both `tests/` and `apps/api/tests/` mirrors):
- `test_workspace_chat_pre_event_stream_failure_returns_actionable_tool_error`
- `test_collect_agent_reply_details_rejects_normal_return_without_finish`
- `test_collect_agent_reply_details_keeps_finish_when_task_then_raises`
- `test_workspace_chat_foreign_conversation_errors_without_forking_or_leaking` (a `conv_...` id owned by a different user/workspace is rejected without forking or leaking that conversation)

Verified fresh in this environment (not just claimed): `python3 -m pytest tests/test_workspace_chat_mcp_2266_2269.py tests/test_langdock_workspace_agent_mcp.py -v` → **18 passed**; `apps/api/tests/test_workspace_chat_mcp_2266_2269.py` mirror → **12 passed**. No hang reproduced.

federicodeponte:
Second independent adversarial review (of commit `ee00d2310`) returned:

```
VERDICT: CHANGES-REQUESTED (completed reply can hang on stalled teardown)
```

Finding: once `finish` is observed, `collect_agent_reply_details` (`apps/api/channels/common.py:885`) awaited the stream task with no bound. If `stream_chat`'s teardown (MCP server cleanup / model-client close, `chat_service.py:2686`-`2760`) hung instead of returning, the already-completed reply would never return to the caller -- a second, subtler way to violate #2266's "a completed run must return its result" guarantee. Reproduced directly: a stream that queues text + `finish` then stalls returned `TIMEOUT` instead of the reply.

Fix in `4b24e80e1`: bound that post-finish await with a 5s internal grace period via `asyncio.wait_for(asyncio.shield(task), timeout=5.0)`. On timeout, log a warning and return the already-collected reply immediately; the existing `finally` block still cancels the lingering task (shield prevents `wait_for`'s own auto-cancel from racing it). Pre-finish race logic and post-finish exception logging are unchanged. Also added a `# TODO(#follow-up)` noting `workspace.chat` accepts `timeout_ms` but does not yet enforce it end-to-end (separate, out-of-scope follow-up).

New regression test (mirrored in both `tests/test_workspace_chat_mcp_2266_2269.py` and `apps/api/tests/test_workspace_chat_mcp_2266_2269.py`): a stream that emits text + `finish` then hangs forever in teardown must still return the completed reply within the grace period, not the full hang duration.

Verification (this environment's Starlette `TestClient` still requires a nonexistent `httpx2` package -- pre-existing, unrelated to this PR, confirmed again with a trivial 2-line FastAPI app):
- 10/10 non-TestClient targeted tests pass in both mirrors (`10 passed, 3 deselected`).
- 500/500 stress-probe iterations (success, pre-finish failure, error parts, caller cancellation, post-finish timeout) with zero leaked tasks.
- `ruff check`, `compileall`, `git diff --check` pass.
- CI on this push: Python lint, Secret scan, Dependency review green so far; Runtime tests (ubuntu/windows), MCP tests, Web lint pending.

Requesting another independent re-review of HEAD `4b24e80e1` before merge.

federicodeponte:
## Independent codex adversarial review — CHANGES REQUIRED

Core direction sound; two blockers before merge:

1. **Race in conversation resume (blocker)** — endpoint validates ownership/existence, then `stream_chat()` resolves+fetches again. If the conversation disappears or changes ownership between those ops, `stream_chat()` falls back to CREATING a conversation under a remapped id — violates #2269 "never silently fork". Fix: fail closed with explicit must-exist semantics inside the conversation service / stream_chat(), not an endpoint preflight. (apps/api/chat_service.py resume path)

2. **Pre-completion persistence suppression (blocker, #2266 over-correction)** — `_persist_side_effect()` (apps/api/chat_service.py ~2503-2508, 2560-2565, 2620-2625) makes tool-transcript persistence non-fatal WHILE the agent is still running. A failed `insert_message(role="tool")` is reported successful with an incomplete durable transcript; a resumed conversation then lacks the tool result and may REPEAT a side-effecting operation. Durable tool-result history should stay fatal (or return an explicit degraded result); only UI tool-card persistence may be best-effort. Also: success may return a `msg_pending_...` id though no assistant row was stored.

3. **Low** — the 655-line suite is duplicated wholesale in apps/api/tests/ and tests/. Keep one canonical copy per repo layout.

Confirmed good: no-id calls create distinct persisted conversations; real-id resume shows history; unknown/foreign conv_ ids fail closed with no leak; finish-then-throw + already-done task exception retrieval covered; owner-scoped lookup not weakened; genuine pre-event failures stay tool errors.

Issue #1201, opened 2026-06-15
main red: e2b install-timeout tests contradict the lifted-clamp code (#1200 cherry vs #1127)


federicodeponte:
## Symptom
On `main` (clean checkout), 2 Runtime-suite tests FAIL:
```
tests/test_e2b_artifact_collection.py::test_long_worker_timeout_extends_install_and_sandbox_lifetime
  assert _install_timeout_for_run(1200) == 900   ->  got 1200
tests/test_e2b_artifact_collection.py::test_sandbox_lifetime_caps_at_e2b_one_hour_limit
  assert _install_timeout_for_run(3600) == 900   ->  got 3600
```
main's Runtime CI will go red once it runs (currently queued). This **blocks every PR** that merges main (incl. #1079 e2b request-timeout fix).

## Root cause
`_install_timeout_for_run` was `max(180, min(ts, 900))` (install capped at **900s**). It was changed to `max(MIN_INSTALL_TIMEOUT_SECONDS, min(ts, MAX_RUN_TIMEOUT_SECONDS))` (install now scales to **MAX_RUN_TIMEOUT=3600**) — the **#1127 "lift 900s clamp"** change.

The **#1200 cherry-pick** (`e79eabc6`, "worker-timeout #1114") shipped the **lifted-clamp code** to main but **kept the old tests asserting `==900`**.

## The unresolved contradiction (needs a lane decision)
PR **#1127** is internally inconsistent on this: it adds `assert e2b_driver._install_timeout_for_run(3600) == 3600` (new) while its `test_sandbox_lifetime_caps_at_e2b_one_hour_limit` still asserts `== 900`. The same call can't be both. So the lane hasn't decided:
- **(A) install caps at 900s** (revert the install side of the clamp-lift; keep run cap at 3600) → fix the *code*, OR
- **(B) install extends to MAX_RUN (3600)** → fix the 2 stale *test* assertions (and recompute the `_sandbox_lifetime_timeout` expectations).

The test name "...extends_install..." + #1127's title suggest (B), but the assertions say (A). **Did not guess-fix** — this is the worker-timeout lane's (#1114/#1127/#1200) design call.

## Recommended
Reconcile #1127 to one behavior (code + all tests agree), then either merge #1127 or push the test-side fix to main. Once main is green, #1079 (e2b request-timeout, currently CONFLICTING only because of this region) re-merges and lands.

cc @federicodeponte (your #1127/#1200) — found during the overnight merge-train run.

[...]

Issue #1183, opened 2026-06-15
[CRITICAL] 204 No-Content routes break app startup under 


federicodeponte:
On `main` @ a71d1b13, several refactored routers combine `from __future__ import annotations`, `status_code=204`, and a `-> None` return annotation. FastAPI 0.111 stores the string `"None"` as the response model in this combination, then asserts that a 204 response must have no body, causing an `AssertionError` at import time and the app refuses to start.

Affected routes found:
- `DELETE /users/{uid}` — `apps/api/routers/auth.py:353`
- `DELETE /auth/tokens/{token_id}` — `apps/api/routers/auth.py:428`
- `DELETE /share/grants/{grant_id}` — `apps/api/routers/share.py:124`
- `DELETE /workers/{worker_id}/alerts/{alert_id}` — `apps/api/routers/worker_telemetry.py:348`
- `DELETE /workers/{worker_id}/feedback/{feedback_id}` — `apps/api/routers/worker_telemetry.py:424`
- `DELETE /workspace/tokens/{token_id}` — `apps/api/routers/workspace.py:908`
- `DELETE /workspace/secrets/{name}` — `apps/api/routers/workspace.py:989`
- `PUT /workspace/settings/{key}` — `apps/api/routers/workspace.py:1039`

I have already applied a local fix: change the return annotation to `-> Response`, add `response_class=Response`, and return `Response(status_code=204)` explicitly. The app now starts and tests pass. The modified files are:
- `apps/api/routers/auth.py`
- `apps/api/routers/share.py`
- `apps/api/routers/worker_telemetry.py`
- `apps/api/routers/workspace.py`

Please review/merge these changes or apply an equivalent fix.

/cc @vivek

federicodeponte:
✅ Fixed and **LIVE on prod**. Merged to engine main (`815cafdd`) via the Tier-A cherry (PR #1200) / cloud security (PR #313), deployed to Railway + Vercel (cloud main `287c4da1`), smoke gate passed all 11 routes, backend tests green. Verified prod serves the new engine; em-dashes confirmed 0 on the live landing. Closing.

PR #2303, opened 2026-07-20
Ship a first cut of workspace + per-worker cost visibility

federicodeponte:
## Summary

Relates to floomhq/workeros-cloud#1201. Live user test: a user seeing 500
runs/day asked twice "who pays for this, doesn't this cost a lot" and wanted
visible costs/limits (he praised Claude Code's statusline showing model +
percent-of-limit). This ships a full cut, not a stub.

- **`GET /workspace/spend`** (new): a purpose-built, read-only readout of
  workspace day/month spend-to-date against the configured daily/monthly
  caps. Reuses the existing `services.run_cost` aggregation functions
  (`_workspace_day_to_date_cost_usd`, `_workspace_month_to_date_cost_usd`,
  `_workspace_daily_spend_cap_usd`, `_workspace_monthly_spend_cap_usd`) -
  the same math that already gates the spend-cap enforcement on run
  dispatch, so the number shown always agrees with what actually blocks a
  run. No mutation.
- **`GET /workers/{id}/spend`** (new): a single worker's month-to-date
  spend + its configured monthly cap (null if uncapped). Reuses
  `_worker_month_to_date_cost_usd` + `_spend_cap_for_config` /
  `get_worker_config_for_run`. 404s for a worker the caller can't see (same
  visibility check as `GET /workers/{id}`).
- **Bug fix in `GET /workspace/settings`**: its existing
  `current_day_spend_usd` / `current_month_spend_usd` mirrors were computed
  by calling `_workspace_day_to_date_cost_usd()` / `_workspace_month_to_date_cost_usd()`
  with no `repos`/`user_id` arguments. Without those, the aggregation always
  fell back to the engine's local sqlite file, which is empty on any
  deployment using a different `Repositories` backend (e.g. Supabase on
  cloud) - so cloud always rendered `$0.00` regardless of real spend. This
  PR passes `repos=repos, user_id=auth.user_id` through, matching the
  pattern already used by the spend-cap enforcement path in
  `run_service._enforce_run_spend_caps`.
- **Settings page**: `ModelDefaults` now renders "`$X of $Y spent this
  month`" (plus a daily figure when a daily cap is set) directly under the
  existing "Monthly spend cap (USD)" field, sourced from the new
  `GET /workspace/spend` endpoint.
- **Worker detail page**: the Overview tab's stat row now includes a
  "Spend (mo)" tile (month-to-date spend, or "spend / cap" when the worker
  has a configured monthly cap), sourced from `GET /workers/{id}/spend` via
  a cache-first `useWorkerSpendQuery` hook. Fails soft (omitted if the
  fetch errors), same as every other stat on that row.
- **Test fix**: `test_797_workspace_defaults_enforcement.py`'s
  `test_settings_returns_current_month_spend` seeded a run at a hardcoded
  day-05 and then asserted `current_day_spend_usd >= 4.25` - a
  date-dependent flake that fails on any day after the 5th of the month
  (confirmed failing on `main` before this PR, unrelated to this change;
  fixed here since it's directly in the code path this PR touches).

## Deferred

- **floomhq/workeros-cloud#1183 ("Not reported" cost/tokens on some
  runs).** Investigated the code path
  (`services/ai_observability.py` + `cost.py`'s
  `resolved_cost_usd_from_transcript`): a run's cost/tokens read
  "Not reported" only when no `{"type":"usage",...}` row exists in its
  transcript at all - correct behavior for genuinely non-agent
  (pure-script) runs that made no LLM calls, since there's nothing to sum.
  Whether the specific repro run (`run_16145da453f5`) is that case, or a
  genuine instrumentation gap, needs looking at the actual run's
  transcript/runner type, which isn't reproducible from this local
  investigation. Deeper root-causing (and possibly changing "no usage data"
  vs "confirmed zero cost" semantics in the persisted `total_cost_usd`)
  is a separate, more involved change and is out of scope for this pass -
  filed as a follow-up rather than rabbit-holed here.

(Per-worker spend UI, originally scoped as a possible defer, shipped in
this PR instead - see the worker detail bullet above.)

## Files touched

- `apps/api/routers/workspace.py` - `GET /workspace/spend` + settings bug fix
- `apps/api/routers/worker_admin.py` - `GET /workers/{id}/spend`
- `apps/web/app/settings/page.tsx` - spend readout in `ModelDefaults`
- `apps/web/app/workers/WorkersCollection.tsx` - "Spend (mo)" stat on worker detail
- `apps/web/lib/api.ts`, `apps/web/lib/types.ts` - client + types
- `apps/web/lib/query/hooks.ts` - `useWorkerSpendQuery` + query key
- `apps/api/tests/test_1201_worker_spend_endpoint.py` (new)
- `apps/api/tests/test_797_workspace_defaults_enforcement.py` - new
  endpoint tests + flaky-date fix
- `apps/web/tests/model-defaults-spend-1201.dom.test.tsx` (new)
- `apps/web/tests/worker-detail-spend-1201.dom.test.tsx` (new)
- `apps/web/tests/model-defaults-797.dom.test.tsx` - mock update for the
  new `api.workspace.getSpend` call
- `docs/API.md` - documents the two new endpoints

## Test plan

- [x] `cd apps/api && python -m pytest tests/test_797_workspace_defaults_enforcement.py tests/test_1201_worker_spend_endpoint.py -q` - 27 passed
- [x] `cd apps/web && npm run lint` - 0 errors (21 pre-existing warnings, none in touched files)
- [x] `cd apps/web && npm run lint:emdash` - clean
- [x] `cd apps/web && npx tsc --noEmit` - no new errors in touched files
- [x] `cd apps/web && npx vitest run` - 204 files / 1015 passed, 1 pre-existing skip
- [x] Confirmed `test_1185_session_storage.py::test_cli_command_panel_uses_session_storage` fails identically on `main` before this change (missing file, unrelated pre-existing issue)
- [x] Ran the full `apps/api` pytest suite twice in parallel (this branch vs. a clean `origin/main` worktree, on this shared/loaded box) to separate real regressions from environmental noise: both runs hit failures/errors at the same percentage markers on the same clusters of tests, confirming the noise is pre-existing/load-induced on this box, not introduced by this diff. Re-ran several of the affected tests in isolation afterward and they passed cleanly.

Generated with Claude Code


federicodeponte:
Merge-queue FAIL resolved. `GET /workspace/spend` now resolves the request active workspace with `_active_workspace_id(request)` and passes that ID into daily and monthly cap lookups. Added a regression test that creates a non-default workspace with caps 21/84 while local-default remains 10/50, then verifies the endpoint returns 21/84 for the active workspace. Merged current Floom main so the head retains #2299 and the current engine commits. Local verification passed: Ruff, bytecode compilation, and 28 related tests. The first Windows pass hit an unrelated current-main workspace-chat timeout after 885 tests passed; rerun on the unchanged head passed: https://github.com/floomhq/floom/actions/runs/29761676968/job/88419133882. All seven remote checks are green on `db76a4a97beee9e38d98f32b2b2ff0e49b060279`.

federicodeponte:
Freshness recheck completed after Floom main advanced.

- Merged current Floom main `8d512f63bac062f34af9d4999e8d18648c39c625` into the existing PR branch without conflicts.
- The spend endpoint still resolves `_active_workspace_id(request)` and passes it to daily and monthly cap lookups.
- The cross-workspace regression still proves that a non-default active workspace receives its own caps.
- All seven remote checks are green on final head `a7f8cef2653b650d5f9376bdb24f612da859027b`.

The original local-default lookup blocker remains resolved on the current-main-containing head. This PR was not merged.

federicodeponte:
Final current-main refresh completed.

- Merged Floom main `a8e4208bea1df57b788dbc3e55fecf977261e643` into the existing branch.
- The active-workspace spend-cap implementation and cross-workspace regression remain intact.
- All seven remote checks are green on head `0b9b9bf5ec3d628a1933ec98b9fee857146fb1d2`.

The original blocker is resolved. This PR was not merged.

PR #2331, opened 2026-07-27
Fix run detail page gutters

### LINKED ISSUE #2330: Fix missing gutters on dedicated run detail pages
## Problem

Authenticated `/runs/[id]` pages are classified as full-bleed because the app shell prefix-matches `/runs`. The detail route then loses the standard horizontal and vertical gutters and its vertical scroll container.

The cloud dashboard chrome mirrors the same classification for `/app/runs/[id]`.

## Expected

- `/runs` remains a full-height collection.
- `/runs/[id]` uses the standard padded, vertically scrollable page shell.
- Loading, error, and completed detail states share the same layout.

## Verified root cause

`apps/web/components/layout/AppShell.tsx` applies `fullBleedCollectionPaths` with prefix matching, so `/runs/<id>` is treated like the collection route.

## Acceptance

- Desktop detail gutters: 24px horizontal and 32px vertical at `sm` and above.
- Mobile detail gutters: 16px horizontal and 24px vertical.
- `/runs` collection remains full-bleed.
- DOM regression coverage for detail, collection, trailing slash, and standalone `/run/[id]`.

federicodeponte:
Closes #2330.

## What changed

- Treat only the dedicated `/runs/[id]` route as a standard page.
- Preserve full-bleed behavior for the `/runs` collection.
- Add DOM regression coverage for detail, collection, trailing-slash, and standalone share routes.

## Root cause

The shell prefix-matched `/runs` for all descendants, so `/runs/[id]` inherited the collection shell: no gutters and `overflow-hidden`.

## Verification

- Focused regression: 4/4 passed.
- Full web suite: 1,023 passed, 1 skipped.
- Web lint: passed with existing warnings.
- Web production build: passed.


federicodeponte:
The required Web lint status is failing in the repository-wide production dependency audit before PR-specific validation. This is reproduced from the unchanged main dependency graph and tracked in #2332. All PR-specific tests, full web tests (1,023 passed), lint, and production build pass locally.

PR #2335, opened 2026-07-30
fix(agent): classify provider overload as capacity so the fallback chain engages

federicodeponte:
## Why

Cross-provider fallback (#2301, #2317) and the 30-minute capacity backoff only engage for `{llm_provider_capacity, llm_rate_limited}`. `llm_provider_capacity` was only produced by quota/billing wording (`_QUOTA_ERROR_RE`) or a 429.

Genuine **server-side overload carries neither**, so it fell through to `llm_provider_error`, which is excluded from the fallback set. A transient platform condition therefore killed the run terminally on the first attempt, with no fallback attempt and only the generic 60s retry delay.

Shapes that were uncovered, each verified against the real classifier:

| Provider | Error | Was | Now |
|---|---|---|---|
| Gemini | 503 `UNAVAILABLE` "The model is overloaded" | `llm_provider_error`, no fallback | `llm_provider_capacity`, fallback |
| Bedrock | 503 ServiceUnavailable / "insufficient capacity" | `llm_provider_error`, no fallback | `llm_provider_capacity`, fallback |
| Anthropic | 529 `overloaded_error` | `llm_provider_error`, no fallback | `llm_provider_capacity`, fallback |
| OpenAI | 500 "The engine is currently overloaded" | `llm_provider_error`, no fallback | `llm_provider_capacity`, fallback |

This matters because the production primary is on-demand Bedrock Claude (`WORKEROS_WORKER_AGENT_MODEL=bedrock/us.anthropic.claude-sonnet-4-6`), whose dominant outage shape is exactly 503/insufficient-capacity rather than a quota string. The fallback chain we shipped did not cover the most likely failure mode of the current primary provider.

## Context

Found while investigating the Jul 18-20 capacity incident. That incident was quota-flavoured (verbatim from `run_logs`: `litellm.MidStreamFallbackError: litellm.RateLimitError: Vertex_ai_betaException - 429 "Your billing account has exceeded its monthly spending cap"`) and is already resolved by #2301 / #2317 plus the AI Studio tier bump. Zero LLM-provider failures platform-wide since Jul 21.

This PR closes the adjacent hole that the same investigation surfaced: the non-quota capacity path, which has never been covered.

## Safety

Classification is gated on a **provider-typed exception marker** (litellm / openai / anthropic / botocore exception reprs, `BedrockException` and friends), not on free text. A worker's own outbound HTTP 503 from a third-party API is therefore never reported as our capacity and never enters the 30-minute platform backoff, which would otherwise mask a real worker-side bug. Two tests pin this, including one where the worker error literally contains a provider hostname.

Ordering is unchanged for everything else: auth, model-not-configured and rate-limit all keep priority. Overload is never the caller's quota, so it stays `llm_provider_capacity` even on user-supplied credentials rather than becoming `llm_quota_exceeded`.

## Also

- Extends the one-shot `chat()` fallback regex (`_PROVIDER_FALLBACK_RE`) with the same capacity signals; it previously only covered auth/quota/rate-limit, so the Gemini fallback key never engaged on overload.
- Records which model actually served a run (`served_model_name` through the fallback chain), so a run rescued by a fallback provider is distinguishable from one served by the primary. Previously only the transition was logged.

## Retry bounds

Unchanged and already correct: 3 attempts, 30m then 60m, capped at 1h, covered by the existing `test_llm_provider_capacity_uses_spaced_bounded_retries`. The observed production retry chain matches it exactly. This PR deliberately does not touch it.

## Tests

10 new tests: classification of all four overload shapes, fallback engagement on each, the verbatim incident string, user-credential handling, non-regression of auth/config/rate-limit/generic-provider, worker-side 503 isolation (2), mid-stream fallback restart proving no partial output leaks and that the served model is reported, and no fallback on application errors.


federicodeponte:
### CI status

All jobs green except **Web lint**, which fails identically on `main` (`npm audit --omit=dev --audit-level=high` flagging `sharp` / `postcss` CVEs). This PR is Python-only and cannot affect it. Reference: `main` run [30468635540](https://github.com/floomhq/floom/actions/runs/30468635540) shows the same single `Web lint` failure with everything else green.

`Runtime tests (windows-latest)` took 4 attempts. Each failure was a different unrelated test with `sqlite3.OperationalError: database is locked`; it is a pre-existing flake at roughly 38% on `main`. Filed as #2337 rather than patched here, to keep this PR focused.

### Reviewer verdict

Reviewed adversarially by Gemini 3.1 Pro (Codex is out of credits until Aug 5). First pass on the design was **DO-NOT-SHIP**, on two substantive grounds:

1. free-text regex is too weak a gate, and
2. a worker's own outbound HTTP 503 could be misread as our platform capacity.

Both were folded in before this PR was opened: capacity classification is now gated on a provider-typed exception marker (Python exception reprs only, so a bare hostname like `status.openai.com` does not match), `try again later` was dropped from the capacity regex entirely, and a second commit adds a structural veto so an exception originating from `requests` / `httpx` / `urllib3` / `aiohttp` is never classified as our capacity regardless of what its message quotes. Three tests pin these cases.

Re-review of the actual diff returned **SHIP**, confirming ordering is correct (auth, model-not-configured and rate-limit all keep precedence) and that the worker-side blast radius is closed.

One known limitation, accepted: `served_model_name` reports the first entry when a single run uses multiple distinct primary models. It is telemetry only and wrapped in a `try/except` so it can never fail a run.

